PR #21, opened 2025-02-02
Integrate Tailwind CSS into the project

RubenJavierBachoAlarcon:
Tell me if I did it correctly or if you want to do it another way

- #20


marcelpanse:
> Tell me if I did it correctly or if you want to do it another way

Can you remove the old css file and add some tailwind classes to the home page instead so we see if it works correctly.

Issue #74, opened 2025-02-07
Bug where you increment a card down the table, that it scrolls back up


RubenJavierBachoAlarcon:

marcelpanse:
This is fixed. However, it feels like a hack though. We can probably fix this by making sure the page doesn't render when you increment a card counter, maybe by pulling the state of the counter into the card component and only update that if you click on an increment button instead of the full list. I'll do some test another time.

PR #75, opened 2025-02-07
Fix scroll position restoration issue in React Table

RubenJavierBachoAlarcon:
This PR addresses the issue of scroll position resetting to 0 upon re-rendering in React Table. The fix involves saving the scroll position in [localStorage] and restoring it when the component mounts. This ensures that the scroll position is maintained even after the component re-renders.

This is problem from React Table https://github.com/TanStack/table/issues/963

- #74


marcelpanse:
I'll merge this in, so it works for now. It feels like a hack though. We can probably fix this by making sure the page doesn't render when you increment a card counter, maybe by pulling the state of the counter into the card component and only update that if you click on an increment button instead of the full list. I'll do some test another time.

PR #103, opened 2025-02-09
fix: verify page & memoize stuff

### LINKED ISSUE #101: Make sure computed values are memoized

amoreaulemay:
Fixes #101

### DIFF
--- a/frontend/src/pages/verify/Verify.tsx
+++ b/frontend/src/pages/verify/Verify.tsx
@@ -1,32 +1,31 @@
-import { login } from '@/lib/Auth.ts'
-import { useEffect } from 'react'
+import { type LoginPromise, useLogin } from '@/lib/hooks/useLogin'
+import { Suspense, use, useEffect } from 'react'
 import { ErrorBoundary } from 'react-error-boundary'
 
-function Verify() {
+interface Props {
+  login: LoginPromise
+}
+
+function Verify({ login }: Props) {
+  const user = use(login)
+
   useEffect(() => {
-    login()
-      .then((user) => {
-        if (user) {
-          console.log('logged in')
-          window.location.href = '/'
-        } else {
-          throw new Error('Error logging in!')
-        }
-      })
-      .catch(console.error)
-  }, [])
+    window.location.href = '/'
+  }, [user])
 
   return <span>Logged in! Redirecting...</span>
 }
 
 function VerifyContainer() {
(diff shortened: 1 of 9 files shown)


frontend/src/pages/verify/Verify.tsx
```diff
@@ -1,32 +1,31 @@
...
-          console.log('logged in')
-          window.location.href = '/'
-        } else {
-          throw new Error('Error logging in!')
-        }
-      })
-      .catch(console.error)
-  }, [])
+    window.location.href = '/'
+  }, [user])
```
marcelpanse, line 14: @amoreaulemay I don't like this pattern, lets not do this. Just looking at this code it's:
1. not clear that the component actually does a Login call when it renders
2. when the user changes it redirects, but it's impossible to see from here when the user would change
3. when the user returns null it redirects?

I don't see what the benefit is of hiding this in a an abstraction versus keeping it simple with the useEffect?

marcelpanse (changes requested):

frontend/src/pages/verify/Verify.tsx
marcelpanse: also, the login function executes only exactly once, so it's pointless to memoize that.

amoreaulemay: In the hook if the user is null then it rejects the promise; triggering the ErrorBoundary. You can't throw an error in use effect, it won't be caught by the errorboundary. So in the effect, user is always defined if it gets there. Memoization is necessary to prevent an infinite call loop (the same reason we use an effect now)

marcelpanse: Lets go back to the original, just 1 simple component <Login /> with a useEffect and a state that shows the error if it fails. Everything in one place, super simple. I don't see the reason to overcomplete this with a component + ErrorBoundary + child component + useLogin + useMemo to do exactly the same thing.

marcelpanse (changes requested):

frontend/src/pages/verify/Verify.tsx
marcelpanse: and I just realized it doesn't matter anyway, there was another issue (which i'm implementing now) to switch to OTP auth so we don't even need this verify page anymore... I'll merge the two.

marcelpanse:
merged into otp-auth

Issue #106, opened 2025-02-09
Implement card details view


marcelpanse:
Card detail page should show a bigger version of the card and all the details we have on the card from the json file (like name, hp, attacks, type, etc, etc). 

The page should also show how many you own and an increment/decrement button. 

For now we can open the detail view if you click on a card from the card list (might change later)

amoreaulemay:
@bleejay @RubenJavierBachoAlarcon we now have two PR for the same issue. @RubenJavierBachoAlarcon was already assigned to this

bleejay:
My bad didn't see this until I posted, happy to close

amoreaulemay:
> My bad didn't see this until I posted, happy to close

I see that @marcelpanse already started a review for #116 so maybe we keep that one. Sorry for the trouble!

PR #116, opened 2025-02-10
Add CardDetail Component and Navigation from Collection

RubenJavierBachoAlarcon:
1. CardDetail Component
Added CardDetail Component: Created a new component to display detailed information about a card.
Integrated FancyCard: Used the FancyCard component to display the card image and details.
Handled State Access: Corrected the way state is accessed from the location object to ensure proper rendering of card details.
2. Navigation from Collection
Added Navigation: Implemented navigation to the CardDetail component when clicking on a card in the collection area.
Passed State: Passed the card details as state to the CardDetail component for rendering.

- #106

### DIFF
--- a/frontend/src/pages/collection/components/Card.tsx
+++ b/frontend/src/pages/collection/components/Card.tsx
@@ -5,6 +5,7 @@ import { UserContext } from '@/lib/context/UserContext'
 import type { Card as CardType } from '@/types'
 import { ID } from 'appwrite'
 import { use, useCallback, useMemo } from 'react'
+import { Link } from 'react-router'
 
 interface Props {
   card: CardType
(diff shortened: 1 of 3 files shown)


frontend/src/pages/collection/components/Card.tsx
```diff
@@ -73,7 +74,9 @@ export function Card({ card }: Props) {
 
   return (
     <div className="group flex w-fit flex-col items-center gap-y-2 rounded-lg border border-gray-700 p-4 shadow-md transition duration-200 hover:shadow-lg">
-      <FancyCard card={card} selected={amountOwned > 0} />
+      <Link viewTransition key={card.card_id} to={`/card/${card.card_id}`} state={{ card }}>
```
marcelpanse, line 77: @RubenJavierBachoAlarcon you shouldn't push the state to the component like this. Instead you can use the :id from the route path in the component and lookup the card from the cardDB in the component itself. The reason is that then you can refresh or deeplink to the page and it will still work. (now it will only work coming from that card-click).

marcelpanse (changes requested):

frontend/src/pages/collection/components/Card.tsx
RubenJavierBachoAlarcon: Okey, i Will fix that sorry

RubenJavierBachoAlarcon: But idk what to do with the request changes

marcelpanse: @RubenJavierBachoAlarcon why did you delete the package.json from the root? We need it so it's like a monorepo and you can run commands from the project root. 

(i'll restore it)

RubenJavierBachoAlarcon: SORRY, idk why happens that, i think it was when i merge the main into this branch

marcelpanse: no problemo

PR #147, opened 2025-02-14
Add Carousel for Mobile Users in ExpansionOverview Component

RubenJavierBachoAlarcon:
This PR introduces a carousel for mobile users in the [ExpansionOverview] component. The goal is to provide a more compact view for mobile users, reducing the need to scroll down to see each pack. Users can still access all the data they need, but in a more convenient format.

Changes:
Added a [Carousel] component that wraps the [GradientCard], [BarChartComponent], and [CompleteProgress] components for mobile users.
Used [useMediaQuery] to detect mobile screen sizes and conditionally render the carousel.
Ensured that the layout for non-mobile users remains unchanged.
Why:
Mobile users previously had to scroll down significantly to view each pack's details. This change makes the interface more user-friendly by allowing them to swipe through the packs in a carousel format.


marcelpanse (changes requested):
@RubenJavierBachoAlarcon I would use the https://ui.shadcn.com/docs/components/carousel instead, it looks more clear what to do because of the arrows and has a nice animation. Also, instead having this if-else based on isMobile() and some duplicate code, I would try to always use the carousel, but just hide the left/right arrows if all fits in the screen. What do you think?

RubenJavierBachoAlarcon:
> @RubenJavierBachoAlarcon I would use the https://ui.shadcn.com/docs/components/carousel instead, it looks more clear what to do because of the arrows and has a nice animation. Also, instead having this if-else based on isMobile() and some duplicate code, I would try to always use the carousel, but just hide the left/right arrows if all fits in the screen. What do you think?

i thought about that, but arrows takes so much space, you think its not clear if we dont use the arrows? i try to copy the apple carousel in mobile 
image
https://www.apple.com/shop/buy-mac
They delete the arrows when its in mobile. If we do it in desktop yes.
And for the duplicate code, you have all the reason. Its better to do it all carousel and if all fits the content then delete the arrows. But i think its better to use the isMobile to hide the arrows. What do you think?

marcelpanse:
@RubenJavierBachoAlarcon: Merged main into this and added some improvements to the responsiveness for tablet/mobile. I'll merge this in like this.

PR #170, opened 2025-02-18
Filtering only one of the linkedCardID

vmdumont93:
The previous code was still showing the two old ambers when filtering. I fixed it.


marcelpanse:
Hi, I also just fixed it at the same time it sees, see my comment on the issue

PR #188, opened 2025-02-20
Visualization types, row and card

Bilhalv:
feature required on issue [185](https://github.com/marcelpanse/tcg-pocket-collection-tracker/issues/185)


marcelpanse (changes requested):
@Bilhalv Thanks!

Some remarks:
- there is a unresolved merge conflict in the types.ts file which breaks the build. 
- you didn't use the formatter that is in the project, so it created a lot of code style changes which makes it hard to review. Please run pnpm lint:fix.
- make sure to run pnpm build before you push to make sure there are no errors like the described above.

Let me know once you fixed it, then I can look at it again.

Bilhalv:
ok! but I'm not sure if I did it right, there is 88 modified files, is that supposed to happen?

marcelpanse:
> ok! but I'm not sure if I did it right, there is 88 modified files, is that supposed to happen?

not sure. can you push it, after the build locally runs?

PR #193, opened 2025-02-21
Set Multiple cards amount

1vcian:
Added the ability to set a custom quantity for multiple cards at once. This feature works with filtered cards, making it easier to mark an entire collection as complete during the initial setup. While a batch database operation would be more efficient for updating multiple cards simultaneously without individual requests, this solution is currently the best approach given the lack of direct database access.
image
image


marcelpanse:
Hi @1vcian this looks really cool and very useful!

I do think it needs a bit more explanation though, otherwise I wouldn't understand what that toggle does and even accidentally overwrite my entire collection.

What about you make it a button "Batch update". And if you click the button it opens a dialog that shows an inline alert "You are about to batch update 121 cards based on your selected filters. Select the amount you like to set for each of the cards below. Beware that this overwrites all the current values of the 121 selected cards!"

Then below you can add an input field for the amount you'd like to set. And finally a button "Batch import".

What do you think?

1vcian:
That sounds like a great idea! I agree that a toggle might not be clear enough and could lead to accidental overwrites. A button with a confirmation dialog makes the action much more intentional and user-friendly. I’ll implement it this way, including the inline alert and input field for setting the amount. Thanks for the suggestion!

marcelpanse:
I love it. I will merge it soon.

PR #221, opened 2025-02-24
Install on device popup and template matching for images

1vcian:
Hey,

I’ve added a non-intrusive banner to install the app, making it easily accessible (implemented with translation support). I haven’t had time to test it on mobile yet, but it should work correctly thanks to the manifest.json.
image

I also started (and mostly finished) a component (@/components/cardTracker.tsx) that accepts images as input and uses OpenCV’s template matching to recognize the images and return an array of cards. It could use some filtering and improvements — I’ll probably optimize it with a ML model later to make it faster, but for now, as an alpha version, it works reasonably well.

Currently, it’s called in collection.tsx, but it’s commented out. You can uncomment it to test, debug, and improve it.

Also, I had to make some dummy declarations to define 'cv' since I’m importing it from a CDN, and I had to temporarily bypass build checks because 'cv' doesn’t exist at build time — integrating it into this project has been tricky so far.

input:
WhatsApp Image 2025-02-22 at 02 39 58

output:
image

Let me know if you run into any issues!

Cheers,Lucian


.idea/shelf/Uncommitted_changes_before_Checkout_at_24_02_25,_19_06_[Changes]/shelved.patch
Comment (marcelpanse) [inline .idea/shelf/Uncommitted_changes_before_Checkout_at_24_02_25,_19_06_[Changes]/shelved.patch:1]: Could you remove these shelve pathes and add it to gitignore? Those shouldn't be in git.

frontend/index.html
```diff
@@ -26,7 +26,8 @@
     <meta name="twitter:image" content="https://www.tcgpocketcollectiontracker.com/images/twitter-card.jpg" />
   </head>
   <body class="dark">
-    <div id="root" aria-label="Main Content Area"></div>
+<!--  <script async src="https://docs.opencv.org/4.x/opencv.js" onload="console.log('OpenCV.js loaded')"></script>-->
```
marcelpanse, line 29: I found some NPM repo that you can use: https://www.npmjs.com/package/@techstark/opencv-js
It also includes the typescript typings. This should make it easier to use.

frontend/src/components/InstallPrompt.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useTranslation } from 'react-i18next'
+
+// Define the type for the beforeinstallprompt event
+interface BeforeInstallPromptEvent extends Event {
+  platforms: string[]
+  userChoice: Promise<{ outcome: 'accepted' | 'dismissed'; platform: string }>
+  prompt(): Promise<void>
+}
+
+const InstallPrompt = () => {
```
marcelpanse, line 12: This shouldn't be visible on desktop screens.
Also, it should not re-appear after closing it, so store it in localstorage to make sure it doesn't keep popping up?

marcelpanse (dismissed):
The branch is not up to date with main which results in a couple of merge conflicts. Make sure to work from an updated main branch. Could you update it and solve the merge conflicts please?

frontend/index.html
1vcian: I actually already tried using that NPM repo, but I ran into some issues during integration

marcelpanse:
@1vcian looks very interesting, I will test and merge it soon.
Great work so far, if you want I can give you access to the repo directly. The benefit is that you can commit directly on a branch, which makes it easier for testing and collaborating. We also have a whatsapp group with the core committers. If you'd like to join let me know and send me your email address.

1vcian:
> @1vcian looks very interesting, I will test and merge it soon. Great work so far, if you want I can give you access to the repo directly. The benefit is that you can commit directly on a branch, which makes it easier for testing and collaborating. We also have a whatsapp group with the core committers. If you'd like to join let me know and send me your email address.
> 
> FYI: I merged the install prompt. It had some bugs which I fixed. I also removed the header icon because I think it's taking too much space, the prompt should be enough to nudge the users to use it, and if they don't want it then lets not bother them again :-).
> 
> I'll take a look at the openCV stuff later.
> 
> Since I made some changes make sure to pull from main before doing anything else.

I’d love to be part of the group, thank you! Feel free to send me an invite — it would definitely make collaborating and testing easier.

About the icon — I initially added it thinking it could be useful for users who might change their mind later. Personally, I find it very handy to have the app directly on the screen without needing to open the browser, remember the URL, or save it to favorites. That said, I totally get the need to avoid being too intrusive, so I’m fine with the change.

By the way, I’ve started working on the card identifier using a different and more efficient approach. The logic I’m developing identifies all possible cards in a photo, then transforms the detected rectangles into hashes and compares them with a pre-hashed card database. By measuring Hamming distance between hashes, it finds the most similar matches — and the algorithm is practically instantaneous. Right now, the only thing left is fine-tuning the rectangle detection, which still needs some calibration.

marcelpanse:
The problem with the icon is that most of the times you click on it, it will actually not work. I read about when chrome throws this event and there are rules like if you discard the prompt, it will not throw the event anymore for like a month (which means clicking on the button won't do anything). Same for when the app is already installed (or previously installed). It even doesn't let you prompt for an install in the first few sessions, only after multiple visits to the site, chrome will allow you to prompt for the install :-/
So, currently, it will only show that overlay you make if chrome sends us the event.

@1vcian I added you as a collaborator, so you can clone this repo directly, branch out and push the branch back. I can add you to the whatsapp group if you send me your phone number, you can drop me an email at marcel.panse@gmail.com

PR #223, opened 2025-02-26
Added card counter to sidebox with UI improvements

vmdumont93:


marcelpanse (changes requested):
The component / logic of the counter is duplicated, could you make it a separate component and use that in both places?

marcelpanse:
@vmdumont93 I like that it is a separated component now. However, that's still a lot of duplicated logic between Card and CardDetail. Why not use Card directly in the CardDetail instead?
If we don't want the "A1-1 - Bulbasaur" title below the card, you can make that a property "hideTitle" on the Card. That way we would have just 1 Card with the logic to increment/decrement it. Otherwise, if there is a bug in that logic we'll have to remember to fix it in two places.

vmdumont93:
@marcelpanse made a lot of sense. I updated it.

PR #238, opened 2025-03-01
[Draft] Recommendation for UI + Questions about .json files

rexwithluv:
This PR is not intended to be merged but serves as a base to raise some questions and show the progress so far.  

- Adjustments in **Header** and **HamburguerMenu** for small screens. This is not a final change but rather a suggestion based on the observation that **Overview** and **Collection** are duplicated.  
- Why are both `card_type` and `evolution_type` stored if they seem to contain redundant information?  
- As I mentioned in the issue, is it possible to modify the JSON files containing card data? I would like to be sure of this before proceeding any further. It’s easy to do with an IDE, but is there a preference to keep a structure readable for non-technical users?  
- Currently, images are retrieved via URL, but if they are stored locally, wouldn't it be better to use them directly? This would allow for future adaptation to change images based on language.  

**NOTE:** I understand this might not be the best way to ask about the current code, but I believe it's much clearer with code examples rather than just issue messages.

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -30,13 +30,14 @@ export function Header() {
         <div className="shrink font-bold pr-4 hidden md:block">TCG Pocket Collection Tracker</div>
         <NavigationMenu className="max-w-full justify-start">
           <NavigationMenuList>
+            <div className="shrink font-bold pr-4 block md:hidden">TCG Pocket Collection Tracker</div>
             <NavigationMenuLink asChild>
-              <Link to="/">
+              <Link to="/" className="hidden sm:block">
                 <Button variant="ghost">{t('overview')}</Button>
               </Link>
             </NavigationMenuLink>
             <NavigationMenuLink asChild>
-              <Link to="/collection">
+              <Link to="/collection" className="hidden sm:block">
                 <Button variant="ghost">{t('collection')}</Button>
               </Link>
             </NavigationMenuLink>
@@ -97,7 +98,7 @@ export function Header() {
           ) : (
             <Dialog open={isLoginDialogOpen} onOpenChange={setIsLoginDialogOpen}>
               <DialogTrigger asChild>
-                <Button>{t('login')}</Button>
+                <Button className="hidden sm:block">{t('login')}</Button>
               </DialogTrigger>
               <DialogContent className="border-2 border-slate-600 shadow-none">
                 <DialogHeader>
--- a/frontend/assets/cards/A1.json
+++ b/frontend/assets/cards/A1.json
@@ -8,7 +8,7 @@
     "image": "https://limitlesstcg.nyc3.cdn.digitaloceanspaces.com/pocket/A1/A1_001_EN.webp",
     "attacks": [
       {
-        "cost": ["Grass", "Colorless"],
+        "cost": ["grass", "colorless"],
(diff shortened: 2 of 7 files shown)


frontend/src/components/Header.tsx
```diff
@@ -30,13 +30,14 @@ export function Header() {
...
+            <div className="shrink font-bold pr-4 block md:hidden">TCG Pocket Collection Tracker</div>
             <NavigationMenuLink asChild>
-              <Link to="/">
+              <Link to="/" className="hidden sm:block">
                 <Button variant="ghost">{t('overview')}</Button>
               </Link>
             </NavigationMenuLink>
             <NavigationMenuLink asChild>
-              <Link to="/collection">
+              <Link to="/collection" className="hidden sm:block">
```
marcelpanse, line 40: Showing the overview/collection links on mobile is on purpose because there are 2 ways to use the app:
1. on desktop - usually used for bulk changes and trading
2. on mobile - for checking which pack to open and quick adding card when swapping between the pokemon app and the site. 

In the case of mobile you want to quickly switch between overview/collection so I don't like having to go through a hamburger menu to do that. The other less important pages for mobile are hidden in the hamburger menu (like profile/trading/etc).
On desktop we have more space so we can show more menu items. 

Also, the login button should always be visible if you are not logged in, to make it clear you have to do that first. So I'd rather not change the menu setup :-)

rexwithluv: I don't use the app from mobile usually, but I resized it on my computer sometime and noticed this, I will revert the changes on that commit soon then! ^^

marcelpanse:
> This PR is not intended to be merged but serves as a base to raise some questions and show the progress so far.
> 
> * Adjustments in **Header** and **HamburguerMenu** for small screens. This is not a final change but rather a suggestion based on the observation that **Overview** and **Collection** are duplicated.
> * Why are both `card_type` and `evolution_type` stored if they seem to contain redundant information?
> * As I mentioned in the issue, is it possible to modify the JSON files containing card data? I would like to be sure of this before proceeding any further. It’s easy to do with an IDE, but is there a preference to keep a structure readable for non-technical users?
> * Currently, images are retrieved via URL, but if they are stored locally, wouldn't it be better to use them directly? This would allow for future adaptation to change images based on language.
> 
> **NOTE:** I understand this might not be the best way to ask about the current code, but I believe it's much clearer with code examples rather than just issue messages.

@rexwithluv regarding the json files. I just created a new scraper that is in the /scripts folder of the repo. We can use this to run to get new cards from expansions or the promo packs. We shouldn't change the json files itself, since the scraper can overwrite them if we need to run it again. 
Any changed we'd like in the json format we can simply adjust in the scraper and run it again now. 
The scraper also downloads the images and store them in the repo so they all get served on our own domain. 

Also, let me know what to do with the PR, do we close it?

rexwithluv:
I've also worked on a script to automatically fetch the cards in Spanish (in Python). Should I include it there? Are these scripts run periodically, or does someone need to run them manually when necessary?

Anyway, I haven't been very active these days, but I'll review everything in detail when I'm back.

Of course, the PR can be closed without any problem :)

marcelpanse:
What is the source you get the Spanish versions from? 
Ideally, I'd like to incorporate this into the current scraper.js script we have so it just works in one command and not use another language.

PR #249, opened 2025-03-03
Added a state variable to disable bulk upload button until the user adjusts

### LINKED ISSUE #230: Batch update card list save button should be disabled until you actively change the card incrementer
Otherwise you might forget and accidentely set all your cards to 0.
It should be 0 and disabled by default. But when you set it to one (or to one and back to 0) then the update button should be enabled.

robertfoster550:
Added a state variable to disable bulk upload button until the user adjusts the amount variable

Resolves #230

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,6 +127,7 @@ dist
 
 # Stores VSCode versions used for testing VSCode extensions
 .vscode-test
+.vscode/
 
 # yarn v2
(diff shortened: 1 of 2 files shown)


.gitignore
robertfoster550, line 1: I can certainly remove the changes to the .gitignore, if desired

Issue #121, opened 2025-02-10
Import cards from spreadsheet


marcelpanse:
Create some way of importing the cards from the spreadsheet

note; that this is handy for now to migrate existing spreadsheet users, it is not very useful in the long run assuming everyone will use the app instead, so no point in making some extensive importer tool.

Export to excel as backup would also be nice though

SpyderHunter03:
OH GOSH YES THIS!!

SpyderHunter03:
I am making progress on this feature.  I have the pages for Import and Export created and the import page can currently take in the spreadsheet and extract out the card name, id, count, and expansion name:

![Image](https://github.com/user-attachments/assets/4805c07c-04e6-40c6-9cd4-8607739e9f07)

Next thing to do is to set all the cards in the database, which I might need some help from @marcelpanse with the library endpoints for those.

marcelpanse:
What lib do you use for parsing the CSV? (I know papaparse is pretty good).

For the DB, you have to do it card by card. You can check the code in the Card.tsx component. Basically you grab the DB from 
```const db = await getDatabase()```

Then you check the ownedCards array to see if you already logged it before. If so, set the correct amount_owned and do an update like 
```
await db.updateDocument(DATABASE_ID, COLLECTION_ID, ownedCard.$id, {
  amount_owned: newAmount,
})
```

If you don't have the card yet, you create a new one with 
```
const newCard = await db.createDocument(DATABASE_ID, COLLECTION_ID, ID.unique(), {
  email: user?.email,
  card_id: cardId,
  amount_owned: newAmount,
})
```

SpyderHunter03:
@marcelpanse I am using the xlsx lib:
https://git.sheetjs.com/sheetjs/sheetjs#

It will allow the user to upload the spreadsheet as an XLSX file, and also export their collection into the same XLSX file (so long as I put a copy of it into the codebase)

SpyderHunter03:
I am going to wait on more updates to this until the current code gets approved and in the application so that I don't have to go over again and make a whole bunch of changes when I go to do the Export.

Also, a possible improvement for the Import could be displaying all the cards and numbers gathered from the spreadsheet and allowing the user to select which cards they want to update so they don't have to update EVERYTHING.  Right now it is a complete overwrite of the current user's collection.

TheCre8r:
Looking forward to this feature.

marcelpanse:
[...]

PR #252, opened 2025-03-04
Import Export Feature

SpyderHunter03:
Export saves a file in CSV format with all your cards and the amount.  Import takes in the CSV file from export and then overwrites all the card values.  The options are in the Account Dropdown area.

This DEFINATELY needs some style polishing, but the core logic of the updating cards and outputting cards works great

#121


marcelpanse:
@SpyderHunter03 I checked it and it looks good. Made a couple of small UX improvements. Shall I go ahead and merge this?

PR #257, opened 2025-03-05
Add a filter to the overview page for number of cards

robertfoster550:
I added a filter to the overview page, which will let the user select if they want to check if they have at least 1 copy of a card, or at least 2 copies.

I also adjusted the various components used on the overview page so that they update with this filter, and show the appropriate percentages and values for each case.

As part of this, I moved the "NumberFilter" component into its own file, and added this to both the overview page and the trade page, where it was previously.

I also added some translations to go with these changes (es and pt translations provided by AI - certainly worth double-checking).

Fullscreen examples with differences between 1 and 2 highlighted:
Screenshot 2025-03-05 at 17 03 10
Screenshot 2025-03-05 at 17 03 24

Mobile example:
Screenshot 2025-03-05 at 17 09 37

### DIFF
--- a/frontend/public/locales/en/number-filter.json
+++ b/frontend/public/locales/en/number-filter.json
@@ -1,0 +1,5 @@
+{
+  "numberCards": "Number of Cards",
+  "maxNum": "Maximum number of cards",
+  "minNum": "Minimum number of cards"
+}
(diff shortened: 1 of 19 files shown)


robertfoster550:
I updated some of the language to show that we are checking different values. Is there anything else I should change? I wasn't sure about the "% chance of getting a new card" when doing the 2 filter, since it now shows the chance of getting a card that the user doesn't have 2 of... not exactly new, but close.

PR #259, opened 2025-03-05
Add Mythical Island themed collection JSONs

PlayInKetchup:
Creates a .json file for each themed collection mission (normal and secret) in the Mythical Island expansion


marcelpanse:
@PlayInKetchup could you update from main and then move them to the frontend/assets/themed-collections folder?
I merged the last PR and moved them to that folder.

PR #284, opened 2025-03-15
Add type definition for rarities

xander-marjoram:
Thanks for making such a cool project!

I was having a browse and thought there might be a good opportunity to use TypeScript to improve the type safety when handling rarities within the app.

A couple of thoughts:
1. I tried to avoid adding entries for every rarity to each object by having separate types like `PickableRarity`, etc, but actually in the end I thought it was probably good practice to have a single type and handle each possible value explicitly.
2. I was wondering why there are two different rarities for promo cards ("" and "Unknown"), do you think this is something that could be simplified? I'm not sure if this is something the scraper would handle, I didn't look at that part too closely.

I'm very open to any suggestions you might have, and if you don't like it at all then that's fair enough 😄

### DIFF
--- a/frontend/src/components/RarityFilter.tsx
+++ b/frontend/src/components/RarityFilter.tsx
@@ -2,12 +2,13 @@ import { Button } from '@/components/ui/button'
 import { Popover, PopoverContent, PopoverTrigger } from '@/components/ui/popover'
 import { ToggleGroup, ToggleGroupItem } from '@/components/ui/toggle-group'
 import useWindowDimensions from '@/lib/hooks/useWindowDimensionsHook.ts'
+import type { PickableRarity } from '@/types'
 import { type FC, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 
 interface Props {
-  rarityFilter: string[]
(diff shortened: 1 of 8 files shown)


marcelpanse:
@all-contributors please add @xander-marjoram  for code

marcelpanse:
> Thanks for making such a cool project!
> 
> I was having a browse and thought there might be a good opportunity to use TypeScript to improve the type safety when handling rarities within the app.
> 
> A couple of thoughts:
> 
> 1. I tried to avoid adding entries for every rarity to each object by having separate types like `PickableRarity`, etc, but actually in the end I thought it was probably good practice to have a single type and handle each possible value explicitly.
> 2. I was wondering why there are two different rarities for promo cards ("" and "Unknown"), do you think this is something that could be simplified? I'm not sure if this is something the scraper would handle, I didn't look at that part too closely.
> 
> I'm very open to any suggestions you might have, and if you don't like it at all then that's fair enough 😄

As far as I can tell the rarity of the cards in the P-A.json is always scraped as "Unknown". Are there any cards with a rarity of ""?

PR #286, opened 2025-03-16
Update translations and JSON files

rexwithluv:
- Added new JSON files in the new "common" directory.
- Fixed fallback language now that the language convention has changed.
- Updated `CardsDB` to follow the new JSON files.
- Removed old JSON files that were no longer in use.
- Edited `scraper.js` to improve multilingual support.

### Notes
- Some elements, such as ability descriptions and card names, are missing due to the complexity of the JSON file structure. I believe I need to carefully review the scraping logic to properly handle these cases.
- Certain game terms do not have direct translations, even within the game itself. I have tried to use the most appropriate Spanish words, but any feedback from Spanish-speaking players is welcome!
- I am also working on scraping Spanish card images and looking for the best way to load them. If anyone wants to help, contributions are welcome! I currently have the script in Python, but I still need to port it to JavaScript. I am sourcing the images from [WikiDex](https://www.wikidex.net/wiki/Juego_de_Cartas_Coleccionables_Pok%C3%A9mon_Pocket).


marcelpanse:
@rexwithluv hi, I see you removed the locale files, maybe to rename them? But I don't see the renamed files in that case (maybe forgot to add them?) - or am I missing something?

rexwithluv:
@marcelpanse I removed the files that were in src/locals because they are now in public/locals, not sure who moved them but it looks like a better place for it to me too. Removing them does not cause any errors because in all the necessary files the files that are in /public are called.

PR #318, opened 2025-04-03
275 checkbox to combine fullart and regular cards

### LINKED ISSUE #275: Add checkbox in overview page to treat normal and full-art cards as the same
The idea is if you want 2 cards of each for deck building purposes, you don't care about if it is 2 normal cards or 2 full art cards (or a combination).

robertfoster550:
Closes #275

Adds a checkbox that allows the user to treat full-art cards like basic cards when counting how many are already owned and the chance of getting a "new" card.

This new filter is called "Deck-building Mode". A tooltip explaining the mode has been added in desktop mode. In mobile it seemed too large and in-the way.

Card counts change when this mode is turned on.

When this filter is turned on, the star and crown rarities are removed from the rarity filter, as they no longer make sense.

Images highlight changes that are made between modes:
Screenshot 2025-04-03 at 17 44 35
Screenshot 2025-04-03 at 17 36 35

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -45,6 +45,7 @@
     "react-hook-form": "^7.54.2",
     "react-responsive": "^10.0.1",
     "react-router": "^7.2.0",
+    "react-tooltip": "^5.28.0",
     "recharts": "^2.15.1",
     "tailwind-merge": "^3.0.2",
     "tailwindcss": "^4.0.15",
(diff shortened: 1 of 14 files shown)


robertfoster550:
I needed something to call it, and went with "Deck-building Mode", but if you have a better idea, I'm fine changing the name.

PR #295, opened 2025-03-23
feature: add all French designs cards

mpierrez:
- Moved English card images to `public/images/en-US/`.
- Created a new `public/images/fr-FR/` directory and added French cards designs.
    - [x] Genetic Apex
    - [x] Mythical Island
    - [x] Space Timesmackdown
    - [x] Triumphant Light
    - [x] Promo-A 
- Updated `FancyCard.tsx` to load images from `public/images/fr-FR/` if available, otherwise fallback to `public/images/en-US/`.

PR #320, opened 2025-04-05
feat: update script to scrap images from a google drive with all cards designs in all languages using Google API

mpierrez:
Here’s the updated version:

---

@marcelpanse I’ve modified the script based on the feedback from the comment [here](https://github.com/marcelpanse/tcg-pocket-collection-tracker/pull/295#issuecomment-2746345732), where it was asked whether it was possible to automatically retrieve images for all languages. After some research, I found a Google Drive that lists all the cards in various languages for the game. However, to download them, interaction with the Google API is required.

The issue is that the Drive isn’t always sorted according to the game order. To address this, I created an Excel file (I thought this was cleaner than doing it directly in the code) where the "old_code" corresponds to the card file name on the Drive. This file name is then mapped to a "new_code" for sorting, as it will be alphabetically ordered.

I’ve commented out the other languages for now, but if they are uncommented and the script is run, it will download the images into the appropriate folders (which will be created if they don’t already exist).

⚠️ Important: I didn’t include "en_US" in the folder IDs in the script because the idea is to keep the scraping for "en_US" in place to preserve the .json files and use English images by default in case new cards are released. This ensures that there will always be a default image available if any images are missing from the drive (or if the drive is deleted).

image

image

Additionally, I created a README file in the scripts folder to explain how to generate the service-account-key.json file needed for Google API authentication. The process is entirely free, but feel free to reach out if you run into any issues with it.

Here are the packages I installed:

- googleapis: Used for authentication and interacting with Google APIs.
- sharp: Used to convert images from PNG to WebP format.
- p-limit: Allows limiting the number of concurrent downloads to prevent overloading the API.
- xlsx: Used to parse the Excel file and retrieve the cards in JSON format for reorganization.


package.json
```diff
@@ -34,10 +34,14 @@
     "@tensorflow/tfjs": "^4.22.0",
     "cheerio": "^1.0.0",
     "fs-extra": "^11.3.0",
+    "googleapis": "^148.0.0",
```
mpierrez, line 37: Allow interaction with Google's API

package.json
```diff
@@ -34,10 +34,14 @@
...
     "i18next": "^24.2.3",
     "i18next-browser-languagedetector": "^8.0.4",
     "i18next-http-backend": "^3.0.2",
     "node-fetch": "^3.3.2",
-    "react-i18next": "^15.4.1"
+    "p-limit": "^6.2.0",
```
mpierrez, line 42: Used to convert image files from PNG to WebP format

package.json
```diff
@@ -34,10 +34,14 @@
...
+    "react-i18next": "^15.4.1",
+    "sharp": "^0.34.0",
```
mpierrez, line 44: Used to convert image files from PNG to WebP format

package.json
```diff
@@ -34,10 +34,14 @@
...
+    "xlsx": "^0.18.5"
```
mpierrez, line 45: Used to parse Excel files and convert them into JSON format, helping to reorganize image filenames.

mpierrez:
I also created a second PR to add the cards (it was generated using this script): https://github.com/marcelpanse/tcg-pocket-collection-tracker/pull/321

PR #323, opened 2025-04-06
Hot fix and and little reafctor

rexwithluv:
After a few somewhat chaotic weeks, I’ve finally been able to get back to the keyboard and dive into writing and reviewing code. I noticed that some functions weren’t working and the translations weren’t translating (at least not the "overview" ones). I’ve fixed that, although I’ve only done minimal testing, so I’m not sure how long it’s been broken xD

Also, there are a few commits showing up in this PR that I could’ve sworn were already merged. In any case, the functions should now be working, and the translations too. Hopefully I can gradually get back into the project and catch up on the new features that have been implemented.

### DIFF
--- a/frontend/src/lib/CardsDB.ts
+++ b/frontend/src/lib/CardsDB.ts
@@ -115,6 +115,8 @@ export const sellableForTokensDictionary: Record<Rarity, number | null> = {
 
 const basicCards: Rarity[] = ['◊', '◊◊', '◊◊◊', '◊◊◊◊']
 
+type CardWithAmount = Card & { amount_owned: number }
+
 interface NrOfCardsOwnedProps {
   ownedCards: CollectionRow[]
   rarityFilter: Rarity[]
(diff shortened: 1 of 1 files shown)


frontend/public/locales/en-US/common/packs.json
```diff
@@ -1,11 +1,11 @@
...
-  "everypack": "Every pack"
+  "pikachu": "Pikachu",
+  "charizard": "Charizard",
+  "mewtwo": "Mewtwo",
+  "dialga": "Dialga",
+  "palkia": "Palkia",
+  "mew": "Mew",
+  "arceus": "Arceus",
+  "shiningrevelry": "Shining Revelry",
+  "every": "Every pack"
```
marcelpanse, line 10: @rexwithluv I don't think this is correct. If I change it, then the labels don't show on the Overview page. This is because they are called "pikachupack" in the A1.json (and the others) file.

marcelpanse (commented):
@rexwithluv regarding your commit history. You should delete the repo and re-clone it from here. Then, instead of using main directly, when you make a change, first branch off of the up-to-date main into a new branch -> make your changes -> make PR of the branch.

The reason is because your commits get squashed into a single commit by the PR into my main branch and that causes your main to not be identical anymore to my main.

If you keep main identical to this main and make a fresh branch for every PR, then you won't have this problem anymore.

frontend/public/locales/en-US/common/packs.json
robertfoster550: So, it seems like you are both kind of right, haha. This change does fix the names on the large cards (which are kind of broken currently) but does break the names elsewhere. The screenshot is from the PR branch:

Screenshot 2025-04-07 at 15 46 16

marcelpanse: In the JSON files, it is named like "pikachupack" see: 
Unless Workspace – A1 json  tcg-pocket-collection-tracker  2025-04-07 at 4 30 25 PM

So lets use that everywhere in the language keys as well.

PR #343, opened 2025-04-19
feat: add cards names in differents languages

mpierrez:
This PR adds a new JSON file at frontend/assets/pokemon_translations.json containing Pokémon names in various languages.

The data is based on the pokemon_species_names.csv file from the [PokeAPI GitHub repository](https://github.com/PokeAPI/pokeapi/blob/master/data/v2/csv/pokemon_species_names.csv).

For now, translations for trainers and items are not included.

https://github.com/user-attachments/assets/ee1c6c73-aed9-4789-90f3-39988bb77479

### DIFF
--- a/frontend/src/components/PokemonCardDetectorComponent.tsx
+++ b/frontend/src/components/PokemonCardDetectorComponent.tsx
@@ -5,6 +5,7 @@ import { Dialog, DialogContent, DialogFooter, DialogHeader, DialogOverlay, Dialo
 import { allCards } from '@/lib/CardsDB'
 import { CollectionContext } from '@/lib/context/CollectionContext'
 import { UserContext } from '@/lib/context/UserContext'
+import { getCardNameByLang } from '@/lib/utils'
 import { CardHashStorageService } from '@/services/CardHashStorageService'
 import { ImageSimilarityService } from '@/services/ImageHashingService'
 import PokemonCardDetectorService, { type DetectionResult } from '@/services/PokemonCardDetectionServices'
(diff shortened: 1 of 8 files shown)


frontend/src/components/PokemonCardDetectorComponent.tsx
marcelpanse, line 1: @mpierrez I'm not sure if the scanner handles multiple languages well. Did you test this in different languages? (by loading incognito so the scan hashes are reset).

frontend/src/components/PokemonCardDetectorComponent.tsx
```diff
@@ -116,7 +117,7 @@ const PokemonCardDetector: FC<PokemonCardDetectorProps> = ({ onDetectionComplete
                   const img = new Image()
 
                   img.onload = () => resolve(url)
-                  img.onerror = () => resolve(`/images/en-US/${card.image?.split('/').at(-1)}`)
+                  img.onerror = () => resolve(`/images/${i18n.language}/${card.image?.split('/').at(-1)}`)
```
marcelpanse, line 120: @mpierrez the onerror should point to en-US in case the lang specific image isn't there.

frontend/src/components/PokemonCardDetectorComponent.tsx
mpierrez: Indeed, I hadn't tested the "Scan" part before.

It's working now, the images are displaying correctly.

image

I was only able to get French images from my app, but since the detection is based on visuals, it should still work for other languages. Actually, in French the names are completely different and it still manages to detect them, so it shouldn't be a big issue.

I also modified the `getRightPathOfImage` function to fall back to the `en-US` folder if the image isn’t found in the original path.

frontend/src/components/PokemonCardDetectorComponent.tsx
mpierrez: Fixed in my commit : https://github.com/marcelpanse/tcg-pocket-collection-tracker/pull/343/commits/c102c9166f792a66d84108273f7e85f3ea600e49

mpierrez:
I’ve modified the abc file to include the Italian Pokémon names as part of this PR https://github.com/marcelpanse/tcg-pocket-collection-tracker/pull/345

It looks like none of them differ from the English ones, but at least it keeps things consistent.

marcelpanse:
@mpierrez thanks, appreciated!

PR #361, opened 2025-04-28
feat: add Italian translation

msilvestro:
Add Italian translation for all translatable text.

Tested locally.

A few issues:
- some text is not translated (e.g. "Scan", "Sign up to view your tradeable cards", "X selected, X uniques owned, X total owned" in the Collection page, etc.) -> they feel a bit off, it would be great to transalate them as well (if I can find some time I can try to understand how to do it)
- there is a small bug that prevents the text to be displayed correctly (it displays the key name and not the translated value):
image
image
As you can see, it says "mewtwo, charizard, pikachu" instead of the correct names (in English should be "Mewtwo pack, Charizard pack, Pikachu pack"). At a quick look, the issue is that the `GradientCard` component receives - for instance - "mewtwo" as key but it should instead receive "mewtwopack". I couldn't find as easy fix - it's probably worth a separate issue.

### DIFF
--- a/frontend/public/locales/it-IT/common/packs.json
+++ b/frontend/public/locales/it-IT/common/packs.json
@@ -1,0 +1,11 @@
+{
+  "pikachupack": "Busta Pikachu",
+  "charizardpack": "Busta Charizard",
+  "mewtwopack": "Busta Mewtwo",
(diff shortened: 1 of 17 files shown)


marcelpanse:
Thank you!

PR #366, opened 2025-04-30
feat: add A3 cards in all languages

mpierrez:
Added the "Celestial Guardians" expansion cards for all languages (FR, ES, IT, PT), 

Except for the following cards : 
- P-A_60 (Exeggcute)
- P-A_61 (Froakie)
- P-A_62 (Farfetch’d) 
 
Because they’re not available on the Drive I scrape the images from. I’ll check regularly to see if they get added.


mpierrez:
I also checked and all the images are properly organized on the site, in the correct order 👍🏻

PR #378, opened 2025-05-03
✨ Visual Enhancements: Expansion Filters with Images + Set Icon in Cards Table

javijec:
This pull request introduces two UI improvements aimed at enhancing the user experience:

1. Expansion Filters (ExpansionsFilter)
Added images for each expansion to make filters more visually recognizable.

Applied a centered and improved layout for better presentation.

Goal: Make expansion filters more intuitive and visually appealing.

2. Cards Table (CardsTable)
Introduced the set icon next to the set name in the table header.

This provides clearer visual context, helping users quickly identify card sets at a glance.


frontend/assets/cards/A3.json
```diff
@@ -5391,7 +5391,7 @@
     "retreat": "N/A",
     "rarity": "◊◊",
     "fullart": "No",
-    "ex": "no",
+    "ex": "yes",
```
marcelpanse, line 5394: Why did you change these? These are not ex cards and the rarities should be on the shinies. Did you overwrite by running the scraper?

marcelpanse (changes requested):
Probably need to revert the chances to the cards (see other comment).

Also, could you add some screenshots of the changes you made? That would make it much easier for me to review.

PR #381, opened 2025-05-04
UI/rework main

javijec:
image
image

New UI


marcelpanse:
@javijec to be honest, I don't think the design is an improvement. You shouldn't have boxes within boxes within boxes, that makes it looks messy. Especially because the boxes have different border radius.

I do like the radial chart, only it should be turned a quarter counter-clockwise so it starts with 0% at the top.

I also understand what you are trying to do, I'll give it some thoughts about how we can improve this. Maybe to make a smaller step towards improvement; you can simply replace the word "pack" with the pack image:
TCG Pocket Collection Tracker 2025-05-04 at 8 51 31 PM

javijec:
I understand your point of view. I think the fact that an application has a lot of text goes against what the UI/UX rules are looking for in terms of user-friendliness. Normally, in boxes like this one. We only read the title of the package and don't see the rest of the text. That's the idea I'm looking for by removing all the text.

marcelpanse:
I agree that a UI with less text is always better UX, however, it should still be very clear what it does without the text. So you should still be able to know, if you see it for the first time, that it is the "pack you should open to have the highest probability to get a new card". If you only display a percentage, then you can't know what it means. 

I do think we can make it a bit shorter by removing this part -> " from among Pikachu pack, Charizard pack, Mewtwo pack. "

javijec:
We can look at how to make the user interface clearer. I think we can improve a lot, but we have to work together. It would be good to find someone who makes designs in Figma.

marcelpanse:
I agree. I can put a post on the forum to see if we can find someone.

PR #395, opened 2025-05-09
Test

javijec:
This is just an example to show you what I mean by replacing the bars with percentages. It is subject to change and is not finished.

image

image


marcelpanse:
Also tagging @aredmondd here, Aiden had some ideas on how to improve this.

aredmondd:
Hey @javijec, thanks for putting this together! I like your idea of making the percentages for each pack more visible. Unfortunately, swapping the bar graphs for static images breaks the card’s visual hierarchy, spacing, and skimmability.

When I think about things being easily readable, there are a few things that make elements stand out to me
1. Color
2. Size
3. Location

We already use color, size, and location to distinguish elements—our colors map directly to each pack’s Pokémon (Orange = Charizard, Yellow = Pikachu, etc.), which is familiar and space-efficient. Sprites from the original games may also confuse new users (who haven't played the original games) and clutter the layout.

Please let me know if you have any other mockups/ideas, as I'd love to keep your ideas within future designs!

javijec:
Sounds good to me. We can talk through other means like Discord. My English isn't the best, so I'd like to work together to improve the website.

PR #400, opened 2025-05-10
feat(gradient-card): add localized pack images and language field

javijec:
This commit introduces localized pack images for different languages and adds a 'language' field to the gradient-card locale files. The GradientCard component now displays the appropriate pack image based on the selected language.

I know translations were discussed in a previous PR. I am uploading this as a proof of concept, it may work.
I got all the images from this Google Drive. 
https://drive.google.com/drive/folders/1_2YURmd7dYnCX3VuDLQobLNSJbAeavpt
There are also images of the logos of each set that we can use to adapt them and other interesting things.


marcelpanse:
localhost dashboard 2025-05-09 at 5 05 33 PM

looks ok to me, @aredmondd what do you think?

marcelpanse:
Localhost 2025-05-10 at 9 05 38 PM

Looks ok to me. @aredmondd what do you think?

aredmondd:
I like this! My only suggestion would be to maybe remove some of the things in the images. For example - you could just have `pikachu` instead of the entire genetic apex pikachu. you could remove the 'themed booster pack' on mythical island... just to remove some of the clutter and cut straight to the chase.

That's my personal preference though. Up to you guys!

javijec:
The big problem with getting the rest of the stuff out is that you have to edit the images for all the languages and sets one by one. These images are the ones that are taken directly from the game.

image

aredmondd:
maybe hold off until you have all the images edited? i have no idea how long it will take - so up to you. might be easier to keep as text for now.

PR #403, opened 2025-05-11
chore(ita): add Celestial Guardians translations

msilvestro:
Add Italian translations for Celestial Guardians set.

Update some other translations.


msilvestro:
@marcelpanse @mpierrez

Issue #250, opened 2025-03-03
Add themed collection mission tracking


PlayInKetchup:
It would be nice to have a way to track themed collection missions for each expansion. Perhaps, adding another tab next to "Trade" called "Themed Collection Missions" that contains a list of expansions to select from. Then within each expansion, there is a list of themed collection missions that show what you have and what you need to complete the mission.

Alternatively or additionally, instead of selecting a specific mission, under the expansion, a list could be generated for all the cards missing for all missions (name/id and amount needed).

All themed collection missions can be found here:
https://www.pokemon-zone.com/themed-collections/

marcelpanse:
Hi, I don't want to make yet another page that sort of does the same. Instead I'd like to add a advanced filters dialog on all pages (overview/collection/trading). In the advanced filters you can filter on everything like rarity, name, type, location, artist, etc. Then one of the things you can filter on is the missions (which should also show you in the dialog which missions you already completed). Then if you filter on a mission, you'll see only the cards from that mission and you can still use the other filters like "owned vs missing". 
If we do it like that, then we can also apply the same filter to each page:
- overview: will show you which pack to open for the selected mission
- collection: will show you which cards are in a certain mission (and show you which you have/missing)
- trading: will show you which cards to trade to finish a certain mission

PlayInKetchup:
It might be nice to show which pack to open for the best chances to get any card missing from all missions. Kind of like the recommendation of which pack to open to get a new card.

jejnaj:
Hey I was looking into picking this up. For scalability reasons, @PlayInKetchup, were you manually creating the jsons of the different missions or was there a way to scrape the info?

PlayInKetchup:
I was manually creating the JSONs. I'm sure there's a way to scrape the info, but I don't have the expertise.

marcelpanse:
finished

PR #414, opened 2025-05-17
Themed collections missions tracker

jejnaj:
My attempt at https://github.com/marcelpanse/tcg-pocket-collection-tracker/issues/250.

Disclaimer: I have little to no experience with frontend, so I based everything about that off of existing code. Having started this two weeks ago, I'm also not up to date on the recent UI changes (beyond rebasing of course) if there's any that would be relevant to the missions. There's definitely some things that could be improved and I also have no clue how this looks on mobile, only desktop.

Added a "Missions" tab to the existing pack filter. Based on @marcelpanse's comment here (https://github.com/marcelpanse/tcg-pocket-collection-tracker/issues/250#issuecomment-2698551603), I wanted to make this feature fairly inconspicuous. I obtained missions by scraping Bulbapedia(https://bulbapedia.bulbagarden.net/wiki/Mission_(TCG_Pocket)#Themed_Collections).

Upon selecting Missions for a given expansion, it shows the missions very similarly to how they appear in-game. A card is highlighted if owned and duplicates can appear. Due to this, I chose to omit the ability to see and increment the number of a card someone has since I show as many as they have if that is applicable. If there are multiple options for an unowned card, the card will show the first option and be greyed out and when you click on it, the options for potential matches appear. If you clicked on an owned card, then the normal CardDetail panel appears (this corresponds with in-game mechanics for collections).

When viewing missions, the ownedFilter can be used to display 'All' missions, 'Owned = completed missions, and 'Missing' = incomplete missions. 

Currently I have implemented nothing in terms of probabilities, this just allows users to see what missions they have completed or not and also shows secret missions, unlike in the game. I hope to do this soon though.

image
image


marcelpanse:
Hi @jejnaj looks cool! Got a couple of questions:

Localhost 2025-05-17 at 8 21 06 PM

- This first mission I have already collected, however it shows under "missing' somehow?
- Same mission also instead of showing just the three cards, it shows all the cards I own including all the duplicates. I would expect it to only show the 3 cards from the mission. 
- I would expect missions under "missing" to always have at least one card greyed out, but I have many where all the cards light up, so I have no idea what to do to finish the mission.
- Seems these folders can be deleted, correct? 
Unless Workspace – Mission tsx  tcg-pocket-collection-tracker  2025-05-17 at 8 27 08 PM

jejnaj:
Thanks for taking a look and reviewing! I'll mark this as draft until I get a chance to debug. I was using a burner account for testing so definitely didn't have the full range of possibilities. I'll look more into these cases and anything else I uncover.

And yep I can probably remove those folders now.

marcelpanse:
@jejnaj Hey, seems to work great now :-)

Two remarks:
1. could we scrape the rewards as well from that site and display them (maybe just in the title of the mission between (brackets)?

2. The missions are a bit too hidden I think, so I played around with the filters UI and settled on showing the pack filter next to the expansion filter so you can see the missions. This is in case of desktop, on mobile there is no space so it is in the filters dialog like you already did.
Localhost 2025-05-20 at 10 34 46 PM


Note: I pushed directly on the PR, so you have to pull that if you want to make changes, or I can merge to main first?

jejnaj:
I like the pack filter change! I had been thinking about adding something about that since it was getting a little tedious during testing to open "All filters" every time.

For the rewards, I scraped them just as a single string since I wasn't sure if/how they'd be used. My only concern for displaying it is that some missions have several different types of rewards (usually emblem tickets, shop tickets, and wonder hourglasses) which may be a little cluttered next to the mission name. That type of UI design is a bit beyond my level I think... I can definitely modify the scraper to grab each reward type separately though.

marcelpanse:
I added the rewards in a tooltip. I had to do some parsing magic to get the rewards properly on new lines. Seems good to me. I'll merge this soon.

Localhost 2025-05-21 at 10 03 09 PM

PR #435, opened 2025-05-29
Add A3a expansion

javijec:


jejnaj:
For the missions, could you revert the jsons for the older expansions and then when running the scraper set expansions at the top to only include 'A3a' (kinda like in the card scraper)? That way it doesn't override the manual changes I needed to make to the previous mission jsons. Also it looks like the site I scrape from hasn't added the missions yet anyways so honestly it could be best to put A3a-missions as 
`[
  {
    "expansionId": "A3a",
    "name": "Coming soon",
    "required_cards": [],
    "reward": "Coming soon"
  }
]`
and I'll update it later.

javijec:
If there is no problem, I will edit it in a while if you don't edit it
before, 😄

marcelpanse:
Thanks for adding this. I'll check it now.

PR #452, opened 2025-06-07
Improve scanning hashes

greg19:
I did some optimizations in the hashing. The time gains were so big, that allowed to increase the hash length by 4 while still being faster. The increased hash length makes the recognition *much* more accurate, and because the memory optimization (they are tightly packed in bit arrays) they should still take about 2 times less space.

The problem for me now is how to trigger recomputation of hashes for the users. I think it would be best to use a different store name and delete the old one (to avoid any complicated checking of hash version), but I couldn't figure out how to do this (I am not very familiar with js).

##### Before 1
image
##### After 1
image
##### Before 2
image
##### After 2
image
##### Before 3
image
##### After 3
image
##### Before 4
image
##### After 4
image

### DIFF
--- a/frontend/src/components/PokemonCardDetectorComponent.tsx
+++ b/frontend/src/components/PokemonCardDetectorComponent.tsx
@@ -24,7 +24,7 @@ interface PokemonCardDetectorProps {
 interface ExtractedCard {
   imageUrl: string
   confidence: number
-  hash?: string
+  hash?: ArrayBuffer
   matchedCard?: {
     id: string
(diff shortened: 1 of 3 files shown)


marcelpanse:
Hey @greg19 that is awesome! To change the database table you can simply change the value here in the `CardHashStorageService.ts`: 
```
  private readonly STORE_NAME = 'cardHashes'
```

That should re-create the hash table for all users.

greg19:
OK, I think I solved this issue

PR #475, opened 2025-06-26
scan confirmation

### LINKED ISSUE #470: Add confirmation popup after card scanning
I believe that it will be very useful for us to have some sort of prompt that lets us know whenever a scan is successful and how many cards have been added to the library, i realized that the cards were not added recently in a scan, I erased the pictures and now my library is not in synchrony.

greg19:
This closes #470

I created the scan confirmation page. I don't know anything about UI, so I'd be thankful for adding proper CSS classes in the `renderIncrementedCard` function. I also don't know why my `h3` looks like regular text.

I'm also wondering what is the purpose of the *Increment amount*. I'm not aware of any use case in which you would want anything other than 1 (especially 0 which it currently allows and I kept for compatibility). I would argue it should be removed, but maybe there is something I'm unaware of. Regardless of that, I was able to simplify it to a `input` removing some hand rolled code.

Also I'd like to take an occasion to raise a topic of removing *Edit matches*, as after #452 I had no problems with wrong recognition. But this is probably for another PR, and might need more testing.

image


marcelpanse (changes requested):
see previous comment

greg19:
> Regarding the edit matches, one feature I do use is click on the card if it's a mismatch to deselect it. Sometimes (specifically when you make screenshots of 10 cards at the same time when you open 10 packs) you get cards partly in the screen, like only the top of the card. In that case, it won't match and I need to deselect it.

I'm talking about the button that allows you to see top5 matches of the extracted image and modify the card to add. They were useful before, when there were errors in recognition, but now I don't get any errors. The select/deselect is of course to stay.

> Usually you want to increment the card. I can think of cases where you'd want to decrement instead. For example, when you do a trade and you screenshot the card you traded away. So it is always plus 1 or minus 1.

Currently it can only allows nonnegative numbers, so you can't decrement a card. Also it applies to all the cards at once, so you would need to scan the same image twice for trade. I don't know why someone would like to increment all the cards from one screenshot by more than 1.

As per trade scanning, I think it would be a cool feature. I think both players after the trade get a screen with a predictable card placement, so there could be a different mode for scanning with some simple heuristic for that. I think changing the current amount modifier to fit this use case would be harder to implement and worse user experience.

greg19:
> Hi Greg! To make it cleaner in the UI I would remove the card previews in the confirmation box since you've already just seen them in the match dialog. I would only put a text confirmation in the confirmation dialog like "4 cards added successfully!". That should be enough. Can you adjust that, then I'll test it further.

image

marcelpanse:
> I'm talking about the button that allows you to see top5 matches of the extracted image and modify the card to add. They were useful before, when there were errors in recognition, but now I don't get any errors. The select/deselect is of course to stay.

Yes, that makes sense

> Currently it can only allows nonnegative numbers, so you can't decrement a card. Also it applies to all the cards at once, so you would need to scan the same image twice for trade. I don't know why someone would like to increment all the cards from one screenshot by more than 1.

Instead of having an input how much you'd like to increment, we could replace it with a simple radio button to choose between increment of decrement.

marcelpanse:
Could you resolve the merge conflicts? Then I'll merge it like this.

PR #477, opened 2025-06-26
feat: add missing translatable strings in french and english locales

nicolasfeyer:
I added missing translatable strings that I found across the frontend in en-US locales and their French translations in fr-FR locales.

I hope I did not miss any strings. 
I wanted to add the rule `react/jsx-no-literals` from [eslint-plugin-react](https://www.npmjs.com/package/eslint-plugin-react) but it does not seem to exist an equivalent with biomejs ([https://github.com/biomejs/biome/discussions/4999](https://github.com/biomejs/biome/discussions/4999))

### DIFF
--- a/frontend/public/locales/fr-FR/scan.json
+++ b/frontend/public/locales/fr-FR/scan.json
@@ -10,12 +10,12 @@
   "loading": "Téléchargement et initialisation du modèle de scan, veuillez patienter... ({{initProgress}})",
   "otherPotentialMatch": "Autres correspondances potentielles :",
   "percentMatch": "match de {{match}}%",
-  "scan": "Scanner",
+  "scan": "Scanneur",
   "selectAll": "Tout sélectionner",
   "selectImages": "Sélectionner des images",
   "selected": "Sélectionné",
   "setIncrement": "Définir l'incrément pour les cartes sélectionnées",
   "showEditMatches": "Afficher l'éditeur de correspondances",
-  "title": "Scanner de cartes Pokémon",
+  "title": "Scanneur de cartes Pokémon",
   "updateSelectedCards": "Mettre à jour les cartes sélectionnées"
 }
--- a/frontend/public/locales/fr-FR/scan.json
+++ b/frontend/public/locales/fr-FR/scan.json
@@ -1,0 +1,21 @@
+{
+  "cancel": "Annuler",
+  "clickToSelect": "Cliquez pour sélectionner",
+  "close": "Fermer",
+  "description": "Vous pouvez téléverser plusieurs images à la fois. Le système scannera chaque image à la recherche de cartes Pokémon et affichera les résultats. Vous pouvez vérifier les correspondances, sélectionner ou désélectionner les cartes détectées, et choisir de combien incrémenter la quantité pour les cartes sélectionnées.",
+  "deselectAll": "Tout désélectionner",
+  "enterAmount": "Saisir la quantité",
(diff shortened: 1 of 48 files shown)


frontend/public/locales/fr-FR/scan.json
```diff
@@ -0,0 +1,21 @@
...
+  "clickToSelect": "Cliquez pour sélectionner",
+  "close": "Fermer",
+  "description": "Vous pouvez téléverser plusieurs images à la fois. Le système scannera chaque image à la recherche de cartes Pokémon et affichera les résultats. Vous pouvez vérifier les correspondances, sélectionner ou désélectionner les cartes détectées, et choisir de combien incrémenter la quantité pour les cartes sélectionnées.",
+  "deselectAll": "Tout désélectionner",
+  "enterAmount": "Saisir la quantité",
+  "extractedCard": "Carte extraite",
+  "hideEditMatches": "Masquer l'éditeur de correspondances",
+  "loading": "Téléchargement et initialisation du modèle de scan, veuillez patienter... ({{initProgress}})",
+  "otherPotentialMatch": "Autres correspondances potentielles :",
+  "percentMatch": "match de {{match}}%",
```
afu-dev, line 12:
```suggestion
  "percentMatch": "correspondance de {{match}}%",
```

frontend/public/locales/fr-FR/scan.json
```diff
@@ -0,0 +1,21 @@
...
+  "scan": "Scanner",
+  "selectAll": "Tout sélectionner",
+  "selectImages": "Sélectionner des images",
+  "selected": "Sélectionné",
+  "setIncrement": "Définir l'incrément pour les cartes sélectionnées",
+  "showEditMatches": "Afficher l'éditeur de correspondances",
+  "title": "Scanner de cartes Pokémon",
```
afu-dev, line 19:
```suggestion
  "title": "Scanneur de cartes Pokémon",
```

afu-dev (commented):
Hello, thanks for adding the french strings, I've suggested two changes :wink:

frontend/public/locales/fr-FR/scan.json
nicolasfeyer: Thanks for the feedback! 

Actually, I first thought of “_correspondance_” because it's the right French word. Unfortunately, the word overflows (see picture).

image

We can either rework the UX to allow longer words, or keep “_match_” in French, which isn't the most appropriate word but is understandable for French speakers.

What do you think ?

frontend/public/locales/fr-FR/scan.json
nicolasfeyer: I agree. "_Scanneur_" looks more like a French word. The _Larousse_ [dictionnary](https://www.larousse.fr/dictionnaires/francais/scanner/71306) validates both spellings

We should also apply this to the line 13: 

`
"scan": "Scanneur",
`

afu-dev: I didn't know "scanner" was in the dict haha, but the official reco is to use "scanneur" because the other one is an "anglicisme".

image

frontend/public/locales/fr-FR/scan.json
afu-dev: Mayne we can go with "{{match}}% compatible", it's the shortest we can do imo. But if this doesn't fit, we can keep the "match" word (maybe "{{match}}% de match"?)

marcelpanse:
Merci!

PR #479, opened 2025-06-27
fix: improve collection/trade page scrolling and add sticky filter bar (fixes #478)

### LINKED ISSUE #478: Collection scrolling could be improved.
## Describe the bug
The collection and trade pages are only a bit higher than the screen and the card table is a box with a virtual scroll. Having the mouse on the cards and scrolling results in the cards moving, but scrolling with the cursor anywhere else results in whole page moving only so slightly and doesn't scroll actual cards.

The bug is also present on mobile, but harder to get as the cards take up the whole width (you have to hold and scroll on the navbar)

## To Reproduce
1. Go to the collection/trade page
2. Scroll up and down with cursor outside the cards area

## Expected behavior
The div around the cards shouldn't have restricted height and expand onto the whole page.

## Screenshots
Image

## Desktop:
 - OS: Linux
 - Browser: Firefox
 - Version: 140.0

## Smartphone:
 - Device: Poco F3
 - OS: Android 13
 - Browser: Firefox
 - Version: 140.0

afu-dev:
Here's my try on the scrolling behavior for the collection and trade page.

The whole page can now scroll the cards, the filter bar in the collection page becomes sticky and, when used, the page scroll to the top.

Height calculation logic has been removed.

I'll add the sticky bar in the trade page now (forgot it!).

### DIFF
--- a/frontend/src/components/CardsTable.tsx
+++ b/frontend/src/components/CardsTable.tsx
@@ -2,7 +2,7 @@ import useWindowDimensions from '@/lib/hooks/useWindowDimensionsHook.ts'
 import type { Card as CardType } from '@/types'
 import { type Row, createColumnHelper, getCoreRowModel, getGroupedRowModel, useReactTable } from '@tanstack/react-table'
 import { useVirtualizer } from '@tanstack/react-virtual'
-import { useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react'
+import { useEffect, useMemo, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import { Card } from './Card.tsx'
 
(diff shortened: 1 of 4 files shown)


afu-dev:
fixes #478 btw.

greg19 (approved):

marcelpanse:
Hi @afu-dev. This doesn't work. The reason we use a virtual table is too load a huge amount of rows in it in a performant way. It needs a fixed height to know how much rows it should load, and then on scroll the virtual table will re-use the rendered divs with different content to create the illusion of a large table. This way you can easily put thousands of cards in it without any performance issues. 
Because you removed the fixed height, the table becomes very large and creates all the rows at once in memory, which will have bad performance on old/mobile devices and will get worse over time since the card database will keep growing. 

Of course, the downside of the virtual table is that it needs the fixed height calculation. However, if you would like to be able to scroll it on the sides of the page as well, you could make the virtual table the full page width, but make sure the card rows are rendered in a boxed div with limited width.

Issue #309, opened 2025-03-28
View trade matches on public page.


marcelpanse:
On the public page the collection of the user is loaded (by friend_id), but if you are logged in, your normal collection is also loaded. 

So we can quite easily now make a button on the collection page called something like “view trade matches” (where the “bulk edit” button normally sits). When clicking the button it should open a dialog that on the left shows you a list of cards that you have and the “friend” is missing and on the right of each card a card of the same rarity that the friend needs and you have duplicate of.

Then you can directly see what to trade :-)

If you are not logged in, then the button should trigger the login dialog instead.

marcelpanse:
![Image](https://github.com/user-attachments/assets/03048ee2-f4ae-4df6-bdad-eaf772c9774d)

marcelpanse:
For reference, this is the first step of a more complex trading system:

- [x] making public page to view a collection (so you don't have to post card lists in the forum)
- [ ] comparing a public page with your logged in page to view possible trades (so you don't manually have to compare)
- [ ] make a button to propose a trade which triggers some email notification (so you don't need to chat on the forum)
- [ ] making a toggle "active trading", if toggled we will promote your public page automatically (so you don't have to post it to the forum). 
- [ ] finally, pro-actively propose trades between "active trading player" (so you don't have to click on the click on the public links)

marcelpanse:
@SpyderHunter03 did you make any progress on this, or anything I can take over / finish for you?

SpyderHunter03:
@marcelpanse I haven't had a moment to work on that since the original attempt.  I can push the branch it is on and let you do the small changes to make it more like the expectation.  I will push it by the end of the night.  I am at work right now.

SpyderHunter03:
@marcelpanse Here is the branch with the changes that I made.  I tried to just merge main into it, but there is a part of the Collection.tsx file that I don't know how to properly resolve the changes, so I ended up just leaving it.

marcelpanse:
Exactly! Was already working on that. It's live now :-)

greg19:
It's really cool!

Now the "Cards your friend needs (and you have)" are the cards that a friend has 0 copies of, right? It would be cool to be able to set the maximum number of copies such that the card is considered "needed".

Also, now the card lists are scrollable:

![Image](https://github.com/user-attachments/assets/aa0570fb-4f1a-4836-82e1-9c3ee7452538)

[...]

PR #480, opened 2025-06-28
trade offer

greg19:
My early steps attempting #309  trade propositions.

My idea was to have a table with columns `trade_id`, `from`, `to` (friend ids) and `card_id`.

* On the trade matches page you can select a subset of cards and send a proposition.
* On a separate page you have all the propositions that involve you and you can select a subset of them and remove (either because you don't want to trade the specific card or the trade happened).

This is the most basic interaction possible, I'd like to hear your thoughts what other interactions should be included. My ideas included:
* Agreeing to a trade
* Disagreeing to a trade (the proposer gets some sort of indicator), maybe with a message?
* Canceling a pending (not accepted) proposition (without any notification)
* Marking trade as done (which automatically increments and decrements involved cards)

Trade matches:
image
Pending trades:
image


marcelpanse:
I would simplify it further and just propose a trade of 1 vs 1 card. 
Because basically I have a enabled trade page, my friend clicks on it, he sees all available options, so the friend clicks 1 card he owns that he wants to trade and selects the card to trade it for - that is the proposal. 

This should result in a db record somewhere and an email to me (the trade page owner). In the email should be a button to decline or accept the trade (resulting in the db status field update). 
Also, the accept/decline should result in an email to the friend that proposed it to inform him.
Both email should contain the proposal and the friendids and nicknames. 

We can show these proposals on your trading page, but even that is optional in my opinion.

---

separate feature, but I also want to show a list of users that set their trading page to active on the homepage, sorted by the the one that did so last. I had some steps outlined in this ticket here --> https://github.com/marcelpanse/tcg-pocket-collection-tracker/issues/309

greg19:
> I would simplify it further and just propose a trade of 1 vs 1 card. Because basically I have a enabled trade page, my friend clicks on it, he sees all available options, so the friend clicks 1 card he owns that he wants to trade and selects the card to trade it for - that is the proposal.

Okay, I see how 1vs1 trades can be simpler. So it should be something like:
```
id
offeror: friend id
offeree: friend id
give: card id
receive: card id
accepted: bool
declined?
finished?
```
Right?

> This should result in a db record somewhere and an email to me (the trade page owner). In the email should be a button to decline or accept the trade (resulting in the db status field update). Also, the accept/decline should result in an email to the friend that proposed it to inform him. Both email should contain the proposal and the friendids and nicknames.
> 
> We can show these proposals on your trading page, but even that is optional in my opinion.

I personally don't want any emails from the site and to be able to do everything from the page. Basically I want to game when I decide to game, and not have it using my mental space when I decide to not game. I agree that emails could be nice for some (probably most) people, and they are a desired feature, but everything should be also doable from the site.

Saying so, I'm also not confident in handling emails, so I can just implement the site changes, and maybe someone can pick up from that.

> separate feature, but I also want to show a list of users that set their trading page to active on the homepage, sorted by the the one that did so last. I had some steps outlined in this ticket here --> #309

Agreed, but I think a more sophisticated recommendation system would be nice. Like taking into account the number of trade matches, and not recommending the same users to everyone to balance and diversify the load (I a little bit afraid that the top people will get tens of proposals and other people will get very little).

marcelpanse:
> accepted: bool
> declined?
> finished?

This could be just one status field (varchar) so we have some more flexibility in the status if we decide to add more later (like awaiting_reponse, etc). 

I'll create the table, then you can do everything in the frontend and i'll just add (optional) emails after that.

marcelpanse:
I settled on this table structure:

table name: trades
columns:
id: int (auto incrementing)
created_at: Date (you don't have to provide - will automatically use now())
updated_at: Date
offering_friend_id: string
receiving_friend_id: string
offer_card_id: string
receiver_card_id: string
status: string

I created the table and I also added the RLS (row level security) rules. I couldn't test it, but if I did it correctly, only the offering OR receiving user should be able to select/update their records. Which means you can do `select * from trading` and you should only get the trades your offered or received.

greg19:
You can make the changes yourself, I guess that will be easier 😅 I can later look at them and give some feedback.

I was thinking about having a question mark tooltip above the trade status column that explained the icons, but I haven't been able to make it work. Also the `updated_at` column is unused now, I think it would be nice to have a UI column that shows it (optimally "17m"/"2d")

The auto increment/decrement after trade is a little bit hard to implement, as it shouldn't be triggered by the other person automatically to the partner, as it isn't safe (and even possible with RLS i guess), and keeping the state of the incrementation for both parties is both non-ideal in the trade row, and hard to neatly package in the UI. I was thinking about having yet another button to increment on your end, but that's starting to be a lot of buttons.

Also the security is sub optimal now, as anyone involved can modify the db row however they wish, but I'm not sure if this can be addressed without having a proper backend. And some malicious person can just start flooding other people trade inboxes. Maybe there should be some "block" feature or rate limit, but again, I'm not sure how it could be done in supabase.

greg19:
Hi @marcelpanse, any plans to move forward with this PR? Maybe you can give me the feedback and I can try to polish it.

marcelpanse:
Hi greg, my apologies, it was a bit hectic with the summer vacation and such. I'll make some time soon to review it and make a list of improvements ideas for you.

marcelpanse:
+ merge main into it and solve conflicts

marcelpanse:
Ok i can add those for you

marcelpanse:
Yes already much better!

- increment -> I feel this isn't the correct word since you will increment one and decrement another. Maybe "Update collection"? I also think it needs a confirmation dialog showing something like "Are you sure you want to add A3-157 Morelull to your collection and remove a copy of A1-227 Bulbasaur? This will also hide the trade from the active trades list."
- after hiding a record it didn't show in the view history list, only after a refresh it appeared.
- when a trade if offered, the page reset and it seems like nothing happens. There should either be a confirmation or a redirect to your trade offers list (so you'll see it there with the status).
- TradeMatches.tsx can be deleted I think?

+ I spent some time updating the UX/visual hierarchy of the pages

greg19:
> increment -> I feel this isn't the correct word since you will increment one and decrement another. Maybe "Update collection"? I also think it needs a confirmation dialog showing something like "Are you sure you want to add A3-157 Morelull to your collection and remove a copy of A1-227 Bulbasaur? This will also hide the trade from the active trades list."

I agree about the wording. The confirmation could be a toast to make the experience smoother not requiring additional clicks.

> after hiding a record it didn't show in the view history list, only after a refresh it appeared.

Yea, I'm aware of that. My react became a spaghetti, I'll try to improve it.

> when a trade if offered, the page reset and it seems like nothing happens. There should either be a confirmation or a redirect to your trade offers list (so you'll see it there with the status).

People could want to make multiple offers at once (I often do that), so I think a toast would be better.

marcelpanse:
@greg19 I think I fixed most of the points discussed above. Let me know if we want to merge this into main or if you want to do some more refactoring first. (or deploy, and then do some refactoring is also possible).

greg19:
@marcelpanse I think I fixed all the issues, so this should be good to merge (but please clean the db table before, so people I tested on are not confused 😅)

marcelpanse:
here we go!

PR #488, opened 2025-07-02
Firefox 3d hover fix

### LINKED ISSUE #426: Firefox: 3D hover effect on cards behaves incorrectly
**Describe the bug**
3D effect of cards when hovering over them is not working as intended on Firefox. 

**To Reproduce**
Steps to reproduce the behaviour:
1. Use Firefox 
2. Go to 'https://tcgpocketcollectiontracker.com/#/collection' 
3. Scroll over cards, moving the mouse around
4. The 3D effect of cards moving while hovering is erratic and not anti-aliased.

**Expected behaviour**
I saw a video from a user with very smooth graphics, so it sparked a doubt on my end. 
Although it's not an essential and urgent issue, I think it detracts from the experience. I waited to test this on another machine, realizing what I was seeing on my PC as a Firefox user was not the intended behaviour and was affecting also the same platform on a different machine. The 3D effect of cards is not working on FF, it's kind of wiggling but not as on other browsers (for example, I used Edge to make a comparison on my own machine and on another PC).

**Video comparison between Firefox (left) and Edge (right)**
https://github.com/user-attachments/assets/408627aa-cecd-4b0d-b0d8-5c3b6eb734a1

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox
 - Version: 138.0.4

**Additional context**
Already tested on another machine, same results. 
Also: when you keep the mouse still it keeps moving at times trying to render what happens on other browsers where the card skews towards the pointer and stays like that.

LandOfDar:
Fixes #426 by adding transform override.
https://github.com/user-attachments/assets/9bdf75bd-ed59-42e0-8fa8-f6dd05049590


LandOfDar:
I think I messed up with the double PR. 
Maybe it wasn't very wise to stay up till late in the morning in order to attempt to fix my github setup. 😥
Sorry if this causes any trouble, everything was working fine earlier. I think it's time to take a break. 😅

LandOfDar:
Closing in favour of #489 which contains only the fix.

PR #490, opened 2025-07-02
add filter cardType

StephenAloche:
Hello,
Here is my PR on adding a filter by card type (trainer) and energy in Collection
I don't know if it was requested but it's present in the TCGP app so I though it will be usefull here as well.

Let me know if it need improvement.


frontend/src/types/index.ts
```diff
@@ -3,7 +3,10 @@ export type ExpansionId = (typeof expansionIds)[number]
 
 const rarities = ['◊', '◊◊', '◊◊◊', '◊◊◊◊', '☆', '☆☆', '☆☆☆', '✵', '✵✵', 'Crown Rare', 'P', ''] as const
 
+export const cardTypes = ['colorless', 'fire', 'water', 'grass', 'lightning', 'psychic', 'fighting', 'darkness', 'metal', 'trainer', ''] as const
```
LandOfDar, line 6: Dragon type is missing here

StephenAloche: oh exact ! I corrected, I also order type to match the card display

LandOfDar:
Don't worry, it's a pleasure to develop stuff for the project, but I'm gonna wait to hear more about the subject in case we're already fine with this.
I just want to play it safe with the content because we know how the developers of the game might be 'skeptical' (or contentious) about the usage of the designs. :sweat_smile:

I'm still quite new to GitHub. 
Just to make sure I’m following best practices: if this PR gets accepted, should I open a separate PR for the updated icons, or would it be better to include them in this one? (Although I'll have to learn to do it this way, ahah). 
Either way, I'll be happy to follow whichever approach is considered the best!

StephenAloche:
> Don't worry, it's a pleasure to develop stuff for the project, but I'm gonna wait to hear more about the subject in case we're already fine with this. I just want to play it safe with the content because we know how the developers of the game might be 'skeptical' (or contentious) about the usage of the designs. 😅
> 
> I'm still quite new to GitHub. Just to make sure I’m following best practices: if this PR gets accepted, should I open a separate PR for the updated icons, or would it be better to include them in this one? (Although I'll have to learn to do it this way, ahah). Either way, I'll be happy to follow whichever approach is considered the best!

I'm also new to contributing on open-source project but from what i've read, you can't add anything on this PR, you could fork my branch to make your dev then make a PR on my repo and then it will update this PR. But the simpliest will be that I update this PR myself with new content OR to merge this and then you can make your changes in your own PR.

LandOfDar:
Yes, I think it will be better if I add to this later so that we can keep it focused on different aspects one at a time, so I'll wait. 

Nice work by the way, I wanted to sort by card type for so long!

marcelpanse:
Hi looks good :-)

I wouldn't make our own images, but we could make the webp versions a lot smaller in size, they are now 640px, but could be resized to like 64px to save 10x the size. 

I don't know about the license part, all the images we use of the cards and expansion logos and such are directly from Pokemon, maybe we could add some attribution in the readme, but I'm not sure what that should look like honestly.

Issue #481, opened 2025-06-29
Failed generating hash & Error incrementing card quantities


LandOfDar:
## Describe the bug
This is probably mainly related to my use case with the localized version (I use the Italian one) but there's also an additional issue when using the English locale that is probably independent of language.
Once updated the model to scan cards, many of the names flash in the “An error occurred!” prompt, saying it failed generating hash for “x” card. 
After changing language it downloads the model and, when it finally seems to work and scan cards correctly, it warns that it failed with prompt: “_Error incrementing card quantities: Error: Error bulk updating collection_”.
I suppose stuff like this happens every time an update is dropped, but since I didn't see anyone raising the issue, I thought about adding it here. Feel free to close it if not relevant or if already acknowledged.

## To Reproduce
Steps to reproduce the behaviour:
1. Go to 'https://tcgpocketcollectiontracker.com/'
2. Select Italian language
3. Click on 'Scan'
4. Download the model
5. Generating hash for 'x' card failed.

  Upon changing language to English
1. Click on 'Scan'
2. Download the model
3. Scan cards
4. Error incrementing card quantities

## Screenshots
![Image](https://github.com/user-attachments/assets/5d070e40-73cd-4b11-b9d4-29b7550ca448) 

![Image](https://github.com/user-attachments/assets/885de151-a0d6-4dd3-bcbe-129ce6bc94af)

## Desktop (please complete the following information):
 - OS: [Windows 11]
 - Browser [Firefox]
 - Version [140.0.2]

marcelpanse:
This could be because I added the updated_at column yesterday. But I didn't update the code accordingly. Have to look into this. Unfortunately I'm out the house today so if anyone else spots the problem please make a pr :)

greg19:
> Upon changing language to English
> 1. Click on 'Scan'
> 2. Download the model
> 3.  Scan cards
> 4.  Error incrementing card quantities

From my testing you also need to select an image containing cards you don't own yet

greg19:
> Steps to reproduce the behaviour:
> 1. Go to 'https://tcgpocketcollectiontracker.com/'
> 2.  Select Italian language
> 3. Click on 'Scan'
> 4. Download the model
> 5. Generating hash for 'x' card failed.

If you use English by default, in order to reproduce this error, you also need to go to developer panel > storage > indexeddb and delete all data in cardHashes2 table.

Shinjimx:
[...]

PR #491, opened 2025-07-03
alternate versions

greg19:
Improved the card database by adding `card_id` to `alternate_versions` (and also calculating `card_id` directly in the scraper). It allowed to improve deckbuilding card merging, and heavily improved the loading time of overview (from about 5s to about 2s on my machine) while the deckbuilding mode is enabled.

### Images
Limitless TCG also renamed their images, so I started renaming them in the scraper, instead of copying the Limitless TCG filename. Also there were places in the app that used the limitless CDN directly, which is unreliable (and not nice). In this PR everything should use this site assets.

#### Missing localized images
IIUC the non-English card images weren't downloaded by any script, but instead committed directly. They are missing for the latest expansion, which explains the first part of #481.

### Equivalent cards bug
Also fixed the wrong assumption in card equivalency check:
```ts
const equivalent = (firstCard: Card, secondCard: Card) => {
  return (
    firstCard.name === secondCard.name &&
    firstCard.attacks.length === secondCard.attacks.length &&
    firstCard.attacks.every((a) => secondCard.attacks.some((atk) => atk.name === a.name))
  )
}
```
Which is false:
image
image
There were many more cards like this, and all should now be counted properly.


marcelpanse:
Thanks @greg19 appreciate all the work! Awesome.

greg19:
I see, will look into that

PR #498, opened 2025-07-05
fix phone scrolling + remove need for desktop scrollbar on collection page

LandOfDar:
- fix scrolling on phone to avoid side-wiggles
- extra offset to remove scrollbars on desktop's collection page since it served no purpose, and we scroll the cards table
  - this extra offset is not considered if mobile device is detected to avoid cutting off content visually on phones

I also tested this on 4K resolution, Android, iOS. 
It should work just fine but of course let me know if any change is needed or feel free to edit later.

Here's a video of a side by side comparison of the behaviour on live website vs my edits on desktop and a screen recording of the wiggling behaviour on phone also addressed in this commit.

https://github.com/user-attachments/assets/5525700d-97f2-4148-a3e1-847501fa8ca2

https://github.com/user-attachments/assets/412c6377-4594-4cc6-bd14-a68bbe5dba5c

### DIFF
--- a/frontend/src/components/CardsTable.tsx
+++ b/frontend/src/components/CardsTable.tsx
@@ -25,8 +25,9 @@ export function CardsTable({ cards, resetScrollTrigger, showStats }: Props) {
       if (scrollRef.current) {
         const headerHeight = (document.querySelector('#header') as HTMLElement | null)?.offsetHeight || 0
         const filterbarHeight = (document.querySelector('#filterbar') as HTMLElement | null)?.offsetHeight || 0
-        const maxHeight = window.innerHeight - headerHeight - filterbarHeight
-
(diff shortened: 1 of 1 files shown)


marcelpanse:
@LandOfDar yes give it a shot :-)
Someone did a PR but it was the wrong approach, see my comment here https://github.com/marcelpanse/tcg-pocket-collection-tracker/pull/479#issuecomment-3014133348

LandOfDar:
> see my comment here [#479 (comment)](https://github.com/marcelpanse/tcg-pocket-collection-tracker/pull/479#issuecomment-3014133348)

Yup! I took it in consideration because at first I did the same and yeah, it became really slow (and basically choked my browser :sweat_smile:), and then I came up with this solution, let me see what I can do.

PR #501, opened 2025-07-07
feat: UX and reset logic in FiltersPanel

LandOfDar:
- added a button to clear all filters at once instead of having to reclick them one by one


LandOfDar:
Great, thanks for the feedback, I was doubtful of the positioning as well, I'll look into it 😁

Issue #499, opened 2025-07-06
Use TCGDex API to dynamically fetch card images instead of storing them locally


mpierrez:
## Is your feature request related to a problem? Please describe.

Currently, images are stored locally (hardcoded) within the project. Over time, managing and updating these images has become time-consuming and cumbersome. It also increases the project’s size unnecessarily.

## Describe the solution you'd like

It would be great to integrate the TCGDex API (https://tcgdex.dev/?ref=api.tcgdex.net) to dynamically fetch card images and data instead of storing them locally. For example, using the API to retrieve sets like the "A1" expansion in English via the endpoint:

https://api.tcgdex.net/v2/en/sets/A1

This would simplify image management, reduce project size, and ensure the images and data are always up to date.

## Describe alternatives you've considered

Continuing to store images locally, which becomes harder to maintain.

Using other external APIs, but TCGDex is well-documented and focused on the relevant card game.

## Additional context

Using the TCGDex API would improve maintainability and scalability for the project, especially as new expansions and cards are released.

mpierrez:
I've also found a better, ready-to-use CDN for images:

For exemple:  https://cdn.pockettrade.app/images/webp/fr/A3b_1_FRA.webp

This might be even faster and easier to integrate than TCGDex for image assets

marcelpanse:
I'd rather not be
1. Depending on another party. If they update, replace, remove their images, then it breaks our app.
2. It's very likely against their terms to use their cdn directly 
3. It's not very nice to leech from another site, which means they'll have to pay for extra egress costs

We currently scrape them with a script, which is easy to update / switch to other databases. But at least, once deployed, can't go down.

I personally don't see the problem with hosting them ourselves.

marcelpanse:
@LandOfDar The scraper.js gets the English version of the card, so any artifacts in that comes from limitlesstcg where we pull them from.

Regarding the other languages, someone else committed them manually, I believe he got them from some google drive. If we have a good source where we can get all these images from (like an API), then go ahead and update scraper.js to fetch those images to store them.

marcelpanse:
@LandOfDar regarding your question, I personally only use the English translations, so I don't care much about the other languages. That is why initially the site was only in English. Thanks to some other contributors like mpierrez they added a translation system and the translation files and images and such. I'm totally fine with that as long as others maintain it and don't have to do anything accept a PR every now and then :-)
[...]

PR #518, opened 2025-07-18
fix: fallback for localized images + cleanup

LandDar:
fix: ensure fallback image path resolves correctly when localized image fails
refactor: clarify variable naming and logging in image path resolution


LandOfDar:
this fixes the issue referenced in https://github.com/marcelpanse/tcg-pocket-collection-tracker/issues/499#issuecomment-3050307638, so now it correctly falls back to _en_ cards

Issue #523, opened 2025-07-21
Add a toggle in the filters panel to search in all card text


marcelpanse:
As it is now, our search function on the collection page only shows results from within card titles. For example, searching for the word “poisoned” shows no card matches, which is a bummer when I’m trying to plan a poison-centric deck. But it would be awesome if the feature searched all text (title, move names, and move effects) to give me a list of cards that met my prompt.

This should not be a default for the search function but a toggle/checkbox next to the field to enable a full card text search in the filters panel.

StephenAloche:
I was thinking at the same functionnality for the same word ^^.
You can assign it to me, i'll check this

However I started to look at how to do it, and from what I see, who only have data in english right ?

marcelpanse:
> However I started to look at how to do it, and from what I see, who only have data in english right ?
Yes, only the names of the Pokemon are translated, but not the other fields.

PR #530, opened 2025-07-24
filter all text

StephenAloche:
Issue : https://github.com/marcelpanse/tcg-pocket-collection-tracker/issues/523

Add a toogle to allow the search input to check all text on cards
card.ability.name
card.ability.effect
card.attacks.name
card. attacks.effect

image

image

image image
translation are done automatically,
I juste translate some elements in french

### DIFF
--- a/frontend/assets/tools_translations.json
+++ b/frontend/assets/tools_translations.json
@@ -177,9 +177,9 @@
 
   "leftovers": {
     "en-US": "Leftovers",
-    "es-ES": "",
-    "pt-BR": "",
+    "es-ES": "Restos",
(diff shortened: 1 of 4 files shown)


StephenAloche:
Sorry for the duplicate, I messed with my branches

marcelpanse:
@StephenAloche could you move the checkbox into the filter panel instead, so that will keep the default UI clean.

StephenAloche:
> @StephenAloche could you move the checkbox into the filter panel instead, so that will keep the default UI clean.

Done,
I saw that the serach input was clear on dialog closing, but can't manage to make it persist

PR #532, opened 2025-07-27
Expansions filter as dropdown

greg19:
As there are more and more expansions, the tabs filter becomes less and less viable, which eventually would need to be addressed. This PR does it by changing it from tabs to dropdown.

image
image
image
image
image


marcelpanse:
@greg19 I added a small fix for the font sizes. The branch needs to be updated though, can you manage the merge conflicts?

PR #534, opened 2025-07-31
Fix card odds calculations to handle A4 baby pokemon

### LINKED ISSUE #533: Fix pull rate calculations to account for new baby pokemon in A4
## Describe the bug
The current code doesn't account for the new way the baby pokemon were introduced. A4 introduces a new pack type different to regular/rare packs being a regular pack + 1 card. This makes a pack with 6 cards where the 6th card is one of the baby pokemon - either a 3 diamond or 1 star card as of these packs.

Also these baby pokemon cannot be found in either normal regular packs or rare packs - only in the new 6 card packs.

The cards asset file needs to be updated to add a baby field so that the calculation functions can be updated to account for this.

## To Reproduce
All accounts will see an incorrect odds rate for packs containing baby pokemon.

## Expected behavior
The odds should be calculated correctly.

gregormaclaine:
Fixes #533

With the introduction of baby pokemon in A4, the way the odds for each card are calculated needed to be updated as there is now a new 6-card pack that has distinct from the regular and rare packs from previous expansions. Baby can strictly be found as the 6th card in a 6-card pack.

This PR updates the scraper.js script to add a "baby" field to each of the cards. Currently the Limitless TCG site doesn't have a baby field, but at the moment all baby pokemon are basic, have 30 health, and have no weaknesses.

The only cards with no weaknesses are dragon-type and babies for now, and there are no dragon-type baby pokemon, so this should be sufficient to distinguish the cards, but the basic and health criteria are added as extra security.

The odds calculator function has been updated to use the new odds if a pack has the new "containsBabies" field equal to true.

Finally a new "Baby: ___" line has been added to the expandable card details pane. However, translations have not been added for non-English languages yet.

### DIFF
--- a/frontend/src/types/index.ts
+++ b/frontend/src/types/index.ts
@@ -35,6 +35,7 @@ export interface Expansion {
   tradeable?: boolean
   promo?: boolean
   containsShinies?: boolean
+  containsBabies?: boolean
 }
 
 export interface Pack {
@@ -67,6 +68,7 @@ export interface Card {
   rarity: Rarity
   fullart: string
   ex: string
+  baby: 'yes' | 'no'
   set_details: string
   pack: string
   alternate_versions: {
--- a/frontend/src/types/index.ts
+++ b/frontend/src/types/index.ts
@@ -68,7 +68,7 @@ export interface Card {
   rarity: Rarity
   fullart: string
   ex: string
-  baby: 'yes' | 'no'
+  baby: boolean
   set_details: string
   pack: string
   alternate_versions: {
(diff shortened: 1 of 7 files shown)


frontend/src/types/index.ts
```diff
@@ -67,6 +68,7 @@ export interface Card {
   rarity: Rarity
   fullart: string
   ex: string
+  baby: 'yes' | 'no'
```
greg19, line 71: I'd prefer boolean

marcelpanse (changes requested):

gregormaclaine:
Made the change, I agree that it's better. I was trying to keep inline with the standard set by the `fullart` and `ex` fields of the same type, which are also 'yes'/'no'.

Issue #134, opened 2025-02-12
Deck builder


marcelpanse:
We have to further spec this feature, but the idea is a page where you can:
1. easily build a deck with your cards
2. store your decks
3. see popular decks (scrape from internet?)

We can combine it with the trade page to put a filter on the trade page to filter by deck and see the missing cards from your deck (and later see which friends have duplicates of them).

RubenJavierBachoAlarcon:
I can go with that, when we have the filter component

RubenJavierBachoAlarcon:
although I had thought, in a section where the best decks are. Because a deck builder is already in the game. Because a deck system where you put likes on the decks and so on, I don't know if users would use it. If there were tournaments or something like that maybe yes. Although this function can be used for other things too.

marcelpanse:
The main benefit would be if you create a deck, then the trading page can automatically show you which cards you need to trade to make your deck (and possible suggest friends that have those cards). 

maybe not the highest priority at the moment

RubenJavierBachoAlarcon:
Okeey, and maybe calculate the probability of the pack to find the deck cards u need?

marcelpanse:
Ah yes, that is also a good idea so you know which pack to open.

StephenAloche:
Hello, I'm working on something, not really a deck builder but more a visualizer of meta deck and wich cards are missing
it look like this for now.
I also add in wich booster each card is missing

<img width="500" height="329" alt="Image" src="https://github.com/user-attachments/assets/e4e6e3dd-7c14-4ff7-98f9-0838380d47a6" /> 

<img width="150" height="326" alt="Image" src="https://github.com/user-attachments/assets/e6ac54ee-ccca-401a-bf2a-239648a9d21e" />

Obviously it still need work but what do you think could it be something usefull ? What data or feature do you think could be missing?


I also have question about the scrapping
I checked Game8, and it might be feasible, even if the data from the DOM would require a lot of processing. Otherwise, there’s PocketTrade, which will be better and it exposes all the data through api, but I’m not sure how “acceptable/legal” it is to scrape that source.

Also, how do you run your other scrapers ? you run it locally ?

PR #546, opened 2025-08-08
meta-deck: [] add deck page + scrapper 2 sites

StephenAloche:
Issue : https://github.com/marcelpanse/tcg-pocket-collection-tracker/issues/134

- Add a Deck page to compare the collection and meta deck from game8.com or pockettrade.app
the first one being currently use  because of reliability.

- Add scrapper to both site that will map deck to a json file 

- Add a search bar to find a deck by it's name

image

image

Later we could imagine having a sort or filter by energy and rank.

Let me know if you want UI adjustments.


marcelpanse:
Hi @StephenAloche, good stuff! Some UX feedback: 

- [ ] clicking on a card should open the card sidebar with the card details, currently it closes the entire deck 
- [ ] only some cards have the expansion logo, seems inconsistent, maybe just remove it in favor of clicking on the card to open the details (which will also show you the alternatives for the same card)
- [ ] alignment list view:
Localhost 2025-08-09 at 9 38 26 AM

PS. get the tooltips from here
Best Decks Tier List (Updated for Wisdom of Sea and Sky) Pokemon TCG Pocket (PTCGP)｜Game8 2025-08-09 at 9 43 35 AM

- [ ] alignment details view
Localhost 2025-08-09 at 9 51 06 AM

StephenAloche:
something like this ?

image

About the expansion logo, I’m not really in favor of removing it. I only kept it so it’s easy to see right away which expansion a card is from, instead of having to click on each card one by one. Some cards don’t have the logo because they’re ones you already own, so it wouldn’t really make sense to show it. But I get you point about inconsistency and  in the end it's your site and it’s totally up to you.

frontend/src/types/index.ts
```diff
@@ -19,6 +19,7 @@ export interface AccountRow {
 }
 
 export interface CollectionRow {
+  englishName: string
```
greg19, line 22: Why do you need to add this field? How is it different from `allCards[row.card_id].name`?

frontend/src/lib/CardsDB.ts
```diff
@@ -78,9 +78,21 @@ export const allCards: Card[] = [...a1Cards, ...a1aCards, ...a2Cards, ...a2aCard
 
 export const allCardsDict: Map<string, Card> = new Map(allCards.map((card) => [card.card_id, card]))
 
+export const allCardsDictName = new Map(
```
greg19, line 81: Different cards can have the same name, we can't do that.

frontend/src/types/index.ts
StephenAloche: see my response of the review on CardsDB.ts
the field is irrelevant and I ofrgot to remove it.

PR #565, opened 2025-08-22
feat(i18n): add localized set images

afu-dev:
I used the Google Drive folder with all the assets datamined to fetch each localized image for each set.

Here's the link to the folder containing all used images: https://drive.google.com/drive/folders/1uSVMnBtCBzDHQtD6vAcuZTnMNmbh9c3i

I had to convert all images with the following command:

```code
$ mogrify -channel A -threshold 300 *.png # make all transparant pixel equally transparant
$ mogrify -trim +repage *.png # remove the transparant border
$ mogrify -format webp *.png # convert to webp
```

Then I choosed the same implementation as the FancyCard component.

All images don't have the same size/ratio, I tried display them the best possible.

<details>
<summary>Here's the look of the implementation</summary>

EN :us:
image

ES :es:
image

FR :fr:
image

IT :it:
image

PT :brazil:
image

</details>


LandDar:
Looking good! I was looking forward to this 😁

PR #570, opened 2025-08-24
feat: add hyphen formatting to friend ID display

afu-dev:
image

(And even in the text copied to the clipboard!)

Note: I can also go with a thin space or a no-break space. In-game, the Friend ID is displayed with hyphens.

### DIFF
--- a/frontend/src/lib/utils.ts
+++ b/frontend/src/lib/utils.ts
@@ -9,6 +9,14 @@ export function cn(...inputs: ClassValue[]) {
   return twMerge(clsx(inputs))
 }
 
+export function formatFriendId(friendId: string): string {
+  if (!friendId || friendId.length !== 16) {
+    return friendId
+  }
+
+  return friendId.match(/.{4}/g)?.join('-') || friendId
+}
+
 export function getCardNameByLang(card: Card, lang: string): string {
   if (card.name === undefined || card.name === null) return ''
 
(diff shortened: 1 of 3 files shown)


marcelpanse:
You could do this:
https://codepen.io/codyastanley/pen/NWKePdm

It adds the hyphens using the ::before pseudo in css, which means it isn't real text so if you select and copy it, it will give you the normal number :-)

afu-dev:
That's a nice solution, but the implementation may be a bit overkill (new component, few function calls...).

How about a small "copy to clipboard" button next to the friend ID?

Example on Pokémon Zone:
image

marcelpanse:
Sure, also fine by me.

afu-dev:
I've created a component for the Friend ID with the possibility to display a copy button. The button put the friend ID, without hyphen, in the clipboard of the user (and display a toast notification saying that it did it successfully).

image

Here's the TradeOffers page with and without the copy button.

image
image

I'm wondering if this feature is worth it or not. It add cluter to the page and doesn't add something that will be actively used (once we added a friend, we don't really need to copy the friend ID later).

greg19:
I think it's nice. Some things to consider:
- removing the parentheses
- scaling the button, now it's much taller than the text on the offers page, which makes it a little bit disturbing
- adding it to `Collection.tsx`, maybe with a similar header like on the `TradeWith.tsx`, the current box is not that nice
- If it is on `TradeWith` and `Collection`, maybe the button isn't necessary on `TradeOffers` to remove the clutter, or maybe skip the trade id at all there?

marcelpanse:
- removed parenthesis
- removed from tradeoffers
- made font-sizes the same in the trading with headline to make it more uniform

PR #573, opened 2025-08-25
feat: added missing translations

afu-dev:
- Added missing translation to the locales (Copy to clipboard, strings in the bar chart on homepage...).
- Used locale Pokémon names in decks, trades pages and export.


marcelpanse:
@afu-dev I just merged a big change in the trading pages. There are probably some more missing translations in those new pages. I'll merge this already, but just so you know

PR #581, opened 2025-08-26
feat: added missing translation (+fr) for the new trade system

afu-dev:
I added the missing string into the locale files. I added all untranslated english strings to the french locale files. Also used the localized Pokémon names for the different places that use it.


afu-dev:
I think I did an oopsi with Git, I'll fix it

PR #593, opened 2025-08-29
Fixed collections not being returned for users without public accounts

JS2K8:
Issue seemed to be caused by commit 8ba7704. Users who don't have a public account seemed to be the only ones affected due to them not technically having an "`account`" but rather just a "`user`" with a linked email. There may be a better solution but this worked for me when running locally at least.


JS2K8:
@marcelpanse That's okay. Glad it's all working again :)

PR #595, opened 2025-08-30
Remove collection reload

### LINKED ISSUE #594: Collection page jumps to top when changing a card amount
@greg19 

When you scroll a bit down and change a card amount, the collection page jumps back up because React rerenders the whole thing. 

This wasn't the case before, but it caused by the refactoring of the cache. I think before it would only update the card itself and not trigger a setOwnCards which will cause a rerender of the entire page. I'm guessing now it will rerender everything, which is makes it very hard to update your cards.

### LINKED ISSUE #596: 400 from supabase when trying to fetch other people collection
Image

The error doesn't say anything, it's just `Object { message: "" }`

Fetching you own collection and other functions work.

greg19:
Fixes #594
Fixes #596 

The fix was simply to remove window.location.reload(). From my testing, this solves the issue, but I don’t know why it was there in the first place. I’d appreciate it if you could confirm that removing it doesn’t break anything.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,19 +1,18 @@
 import loadable from '@loadable/component'
 import { useEffect, useMemo, useState } from 'react'
 import { ErrorBoundary } from 'react-error-boundary'
-import { Route, Routes, useLocation } from 'react-router'
+import { createHashRouter, Outlet, RouterProvider, useLocation } from 'react-router'
 import DonationPopup from '@/components/DonationPopup.tsx'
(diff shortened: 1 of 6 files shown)


greg19:
Ah, I see that it will probably break when switching from friend collection to your own.

greg19:
Should be good now. I used react router loader functions to simplify the flow a little bit.

marcelpanse:
No worries, it's fine to go fast as long as we also fix bugs quickly. I also didn't think of the fact that (most) users don't have an account record.

PR #602, opened 2025-08-31
Update local collection state atomically

### LINKED ISSUE #601: Bug: collection didn't update properly
Hello,

Just went through my Lugia collection and updated it one by one after opening a 10-packs. Sadly, half of them didn't update. Went though it again, again half of the remaining didn't update.

I'm on a Firefox based browser (Zen). No error in the console, just the "cache collection update".

I'll try again later today.

greg19:
Might fix #601, but needs confirmation.

I believe the bug is related to creating a copy of whole `ownedCards` and holding on it for quite amount of time (basically two db detours). Other running updates (that were started earlier, but not yet commited their changes to `ownedCards`) might modify it in the meantime, and the later update reverts the state because it made the copy before the commit. Turns out concurrency is hard.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -51,17 +51,6 @@ function App() {
     const nowString = now.toISOString()
     const rows: CollectionRow[] = rowsToUpdate.map((row) => ({ ...row, email: user.user.email as string, updated_at: nowString }))
 
-    const ownedCardsCopy = [...ownedCards]
-    for (const row of rows) {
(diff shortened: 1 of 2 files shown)


marcelpanse:
@greg19 seems very unstable: https://share.zight.com/BluAdOnq

marcelpanse:
You can wait for the ownedCards to be updated because that happens in React with some latency. 
Last commit fixes the issue: https://share.zight.com/9ZupDXev

PR #612, opened 2025-09-05
[WIP] german localization

mkslzk:
Hello, 
I asked several weeks ago if I could contribute with a german localization. 
Here is the first little part. I hope this fits your style and code of conduct since this is the first pull request.

Thanks in advance! 
Mike


marcelpanse:
Nice! Instead of adding [WIP] in your title, you can mark the PR as draft, which will actually prevent us from accidentally merging it. When you are ready you just mark it as ready.

PR #617, opened 2025-09-06
Total amount

greg19:
image

Added a "Total amount" to card side panel. Also included `ownedCardsMap` in the context to speed up all the `find` operations.

Tested:
- [x] Total amount works
- [ ] `OwnedCardsMap` doesn't break `PokemonCardDetector`
- [ ] `OwnedCardsMap` doesn't break import/export
- [x] `OwnedCardsMap` doesn't break `TradeListRow`


greg19:
This can wait for the react query

afu-dev:
Indeed having the rarity + card_id is a great idea :+1:

PR #618, opened 2025-09-06
React query

marcelpanse:


frontend/src/services/account/useAccount.ts
```diff
@@ -0,0 +1,44 @@
...
+import type { AccountRow } from '@/types'
+import { accountService } from './accountService'
+
+export function useAccount() {
+  const { data: user } = useUser()
+  const email = user?.user.email
+
+  return useQuery({
+    queryKey: ['account', email],
+    queryFn: () => accountService.fetchAccount(email || ''),
```
greg19, line 14:
```suggestion
    queryFn: () => accountService.fetchAccount(email as string),
```

I believe this was because of types, but it was already enforced in runtime by `enabled`, right? Is so, this cast should make it clear.

frontend/src/components/BatchUpdateDialog.tsx
```diff
@@ -16,15 +17,22 @@ interface BatchUpdateDialogProps {
...
+  const updateCardsMutation = useUpdateCards()
+  const { data: user } = useUser()
 
   const onBatchUpdate = async (cardIds: string[], amount: number) => {
-    await updateCards(cardIds.map((card_id) => ({ card_id, amount_owned: amount })))
+    if (!user?.user.email) {
+      throw new Error('BatchUpdateDialog.tsx:onBatchUpdate: User not logged in')
+    }
+    updateCardsMutation.mutate({
+      email: user.user.email,
```
greg19, line 33: Is there a way we can get `user.user.email` inside mutate? This would make the code cleaner

frontend/src/components/EditProfile.tsx
```diff
@@ -50,18 +47,13 @@ const EditProfile: FC<Props> = ({ account, setAccount, isProfileDialogOpen, setI
...
-          friend_id: values.friend_id,
-          is_public: values.is_public,
-        })
-        .select()
-        .single()
-
-      setAccount(account.data as AccountRow)
+      // Update account using mutation
+      await updateAccountMutation.mutateAsync({
+        email: user?.user.email || '',
```
greg19, line 52: Can the user actually be `null` here? If so, I'd add an early return just above `onSubmit`, if not, I think a cast to string (instead of `|| ''`) will signal to the reader that this is the case.

frontend/src/services/collection/collectionService.ts
```diff
@@ -0,0 +1,195 @@
+import { supabase } from '@/lib/supabase'
+import type { AccountRow, CollectionRow, CollectionRowUpdate } from '@/types'
+
+const COLLECTION_CACHE_KEY = 'tcg_collection_cache'
+const COLLECTION_TIMESTAMP_KEY = 'tcg_collection_timestamp'
+const PAGE_SIZE = 500
+
+export const fetchCollection = async (email: string, collectionLastUpdated?: Date) => {
```
greg19, line 8: I would rename it so it's clear that is also might use the cache

frontend/src/services/collection/useCollection.ts
```diff
@@ -0,0 +1,57 @@
...
+import { fetchCollection, fetchPublicCollection, updateCards } from '@/services/collection/collectionService.ts'
+import type { CollectionRowUpdate } from '@/types'
+
+export function useCollection() {
+  const { data: user } = useUser()
+  const email = user?.user.email
+  const { data: account } = useAccount()
+  const collectionLastUpdated = account?.collection_last_updated
+
+  return useQuery({
```
greg19, line 15: The default `staleTime` (which is set to 1m) applies here, so if someone edits their collection on another device it can be stale. It can be unintuitive if someone is really using two devices at the same time (I don't know for what exactly, but it may happen). Maybe consider lowering it, as the cache will be used almost always. I don't know if it wouldn't cause unnecessary react redraws tho.

marcelpanse:
> ### Async mutations?
> I see that in `EditProfile.tsx` a `mutateAsync` is used and everywhere else `mutate` is used. Is `mutate` blocking, or returns immediately and happens in the background? Currently the functions that call `mutate` are still async but don't await anything.

Yes mutate doesn't return anything and happens in the background, but you can bind to updateAccountMutation.isError or updateAccountMutation.isPending etc directly in the React UI and show the correct state there. 
Because of this you can't try/catch the mutate, unless you use mutateAsync, which will block. 

However, I think I should move all the toasts into the mutate function itself so we don't clutter the code with them and it's easier to handle the errors there.

marcelpanse:
> ### Default value for `ownedCards`
> I see a common pattern:
> 
> ```js
> const { data: ownedCards = [] } = useCollection()
> ```
> 
> If I understand correctly, the `[]` is used when `ownedCards` is `undefined`, when exactly can that happen? I can think of:
> 
> * when the user is not logged in
> * when the database errors
> * when we are waiting for the data from the DB?
> 
> Is that correct?
> 
> I think in all of these places we should display something meaningful instead of pretending that everything is correct, but the user just has an empty collection. Silently doing that can be dangerous as we recently learned: [[Closed][#593] My whole collection disappeared](https://community.tcgpocketcollectiontracker.com/t/closed-593-my-whole-collection-disappeared/6717).

In react-query there is a option placeholderData, which I wanted to use at first, it's the initial data the query will return. However, it doesn't return anything if the query's enabled property is false. So if you are not logged in, the query won't run due to the enabled prop, but then it also doesn't return anything so the value would be null. Since all our code depends on it being an array I initialize it like the code above, otherwise we'll have to introduce a lot of null checks.
Note, when you are not logged in and browse the site, it will also stay an empty array until you log in, but you can still view the site and browse the collection and stuff.

frontend/src/pages/trade/components/TradeList.tsx
```diff
@@ -1,32 +1,34 @@
...
   const { t } = useTranslation('trade-matches')
   const { toast } = useToast()
 
+  const { data: account } = useAccount()
+  const { data: ownedCards = [] } = useCollection()
+  const updateCardsMutation = useUpdateCards()
+
   function interesting(row: TradeRow) {
-    return (row.offering_friend_id === account?.friend_id && !row.offerer_ended) || (row.receiving_friend_id === account?.friend_id && !row.receiver_ended)
+    return (row.offering_friend_id === friendId && !row.offerer_ended) || (row.receiving_friend_id === friendId && !row.receiver_ended)
```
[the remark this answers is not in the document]
marcelpanse: Hmm, when I use the account?.friend_id I get identical lists for all my friends in the offers list:

Localhost 2025-09-09 at 10 05 38 PM

But when I use the friendId it actually renders correctly:

Localhost 2025-09-09 at 10 06 47 PM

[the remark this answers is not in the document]
marcelpanse: actually, it doesn't exactly render like on production, so something is wrong here but I'm unsure what is going on:

Prod:

TCG Pocket Collection Tracker 2025-09-09 at 10 08 25 PM

frontend/src/services/collection/collectionService.ts
marcelpanse: The rest of the code just wants to "fetchCollection", it isn't interested in if there is a cache behind it or not, so I think this is fine.

frontend/src/services/collection/useCollection.ts
marcelpanse: I'll set it to 10 sec in this case then. It will re-check our internal cache in that case. 
If you leave it at 1 minute, it means it won't even call our fetchCollection function for that minute until it expires, then it would call the fetchCollection function. So your mobile changes won't be visible for a minute on the desktop - unless you refresh your page, then you'll see it, which I assume most people would do. 

I don't think a low stale time will cause a rerender automatically, it just marks it as stale for the next it gets called (when the component rerenders for whatever reason like switching tabs, or if you use the refresh() function on the useCollection).

greg19:
> In react-query there is a option placeholderData, which I wanted to use at first, it's the initial data the query will return. However, it doesn't return anything if the query's enabled property is false. So if you are not logged in, the query won't run due to the enabled prop, but then it also doesn't return anything so the value would be null.

Got it, thanks.

> Since all our code depends on it being an array I initialize it like the code above, otherwise we'll have to introduce a lot of null checks.

If the type is `CollectionRow[] | null` and you do an early return `if (ownedCards === null) return null` typescript is smart enough to narrow the type to just `CollectionRow[]` afterwards. That's like two extra lines of code, we already do it with `account` (if you merge the if it's zero extra lines of code).

> Note, when you are not logged in and browse the site, it will also stay an empty array until you log in, but you can still view the site and browse the collection and stuff.

Yea, each place would be handled one by one. I agree that showing what would the functionality will look like when you log in is very desired instead of just showing "register to use the site" everywhere.

- The collection, missions and decks need to be usable even when you are not logged it. It would be cool to check that and don't fade missing (all) cards but show them in their normal form if the user is not logged in.
- Export and import readers should definitely check if the user is logged in and disable themself
- `TradeListRow` probably can throw
- `TradeWith` currently would show "No trades available between you and your friend", it would be more appropriate to instead show a message to log in and have two buttons to log in or navigate to friend collection.
- `CardDetectorComponent` should allow you to scan and view matches, but have the button to update collection disabled (with a caption telling you to log in).

I think we can keep it as is in this PR to not make it even longer and address all with separate PRs.

frontend/src/pages/trade/components/TradeList.tsx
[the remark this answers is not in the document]
marcelpanse: Alright, should be solved now in the latest commit

marcelpanse:
@afu-dev fixed the issues you flagged in the latest commit 👍

PR #627, opened 2025-09-13
Improve card tables in trading

greg19:
Closes #471

This adds a reusable `CardLine` component that can be used for lists of cards. It adds a button to open a card details side panel, which solves the issue I raised in #471.

Trade offers:
image

Trade with:
image

There were too little space for the button on narrow screens, so I changed how the trade offers are displayed (on small screens):

image

To make sure you know which card you are giving and which receiving, I also added how the card amounts will change

Issue #622, opened 2025-09-12
Navbar overflows on small screens


greg19:
![Image](https://github.com/user-attachments/assets/4b7342ff-0e8d-4a3b-98f3-d99af672e4db)

PR #629, opened 2025-09-13
Move scan to a separate page

greg19:
As I suggested in #622, this moves the scanning behavior to its own page. This gets rid of the awkward button on narrow screens. Also it's just cleaner to have it as a separate page: there is more space, and we don't have to worry about some unexpected collection modifications on other pages.

I had some other improvements in mind (like confirmation list with the amount changes via `CardLine` from #627, or revamp of the scan match visuals which currently stand out from the rest of the site), but wanted to get this out first so the PRs aren't that big.

I also merged the navbar and hamburger menu translations, as the hamburger was a subset of navbar. And removed the old "offers" translation from very early stages of trade offers.

Tested:
- Links in navbar and hamburger menu work
- Scanning works, including all the state transitions
- Translations work (both on the new page and on navbar/hamburger)


marcelpanse:
Some feedback after testing: 

1. the scan page without the boundaries of the modal look a bit weird, like everything is floating around. It needs some sort of box containing it. 
2. when you click on a card the whole site seems to flash.

See: https://share.zight.com/Z4u4n9lD

3. on mobile pages if you don't have a trade notification, we only show the collection now, which is also a waste of space. Since you use the scan mostly from mobile (open pack -> screenshot -> open site -> scan). I think the scan should always be visible. There is enough space with my latest change even with the trade notification visible and it will saves you a click to open the side-menu to go to the scan page.

greg19:
> on mobile pages if you don't have a trade notification, we only show the collection now, which is also a waste of space. Since you use the scan mostly from mobile (open pack -> screenshot -> open site -> scan). I think the scan should always be visible. There is enough space with my latest change even with the trade notification visible and it will saves you a click to open the side-menu to go to the scan page.

On the French locale you currently might get an overflow at 461px, which is far wider than a typical phone (quick google search says 360-414px, and we should care about the lower spectrum, so we are like 27% too wide):
image
On said 360px it looks terrible on production now:
image


So I don't agree we **always** show the scan button – you just can't accommodate all collection/overview, trade with notification and scan, and I think trade notification is more important than scan button. What do you think of showing scan button only when there are no trade notifications, and showing trade instead when there are some?

marcelpanse:
Yes lets do that. Hide the scan button only if you have a trade notification. 

> > on mobile pages if you don't have a trade notification, we only show the collection now, which is also a waste of space. Since you use the scan mostly from mobile (open pack -> screenshot -> open site -> scan). I think the scan should always be visible. There is enough space with my latest change even with the trade notification visible and it will saves you a click to open the side-menu to go to the scan page.
> 
> On the French locale you currently might get an overflow at 461px, which is far wider than a typical phone (quick google search says 360-414px, and we should care about the lower spectrum, so we are like 27% too wide):
> image
> On said 360px it looks terrible on production now:
> image
> 
> 
> So I don't agree we **always** show the scan button – you just can't accommodate all collection/overview, trade with notification and scan, and I think trade notification is more important than scan button. What do you think of showing scan button only when there are no trade notifications, and showing trade instead when there are some?

PR #638, opened 2025-09-15
Decouple collection and missions

greg19:
Currently missions are considered as packs and deeply coupled with you collection in `Collection.tsx` and `FiltersPanel.tsx` (which does `filteredCards = []` when you choose the "mission pack", which is, lest say, unconventional). This has some non-obvious consequences:

- When viewing missions, you still have all the filters panel available, but it mostly doesn't work. For example, you can change the mission rarity/energy type/maximum number of cards owned, or search for a text, but these doesn't influence the missions shown in any way. This might be confusing for the user who expects to find the 5 charizard mission by typing the card name in the search box.
- You can switch to missions in friends collection. But it actually shows your missions, not friends. It shouldn't behave this way.
- More on developer side: it's hard to add new filters when the `FiltersPanel.tsx` holds so much responsibility not related to actually doing what it is designed to (filtering cards).

This PR makes Collection and Mission a separate thing. You still access the missions in `/collection`, but the component is different and you can switch back and forth. The missions show only relevant filters. And also you can no longer view your missions when on friends collection.

Positioning the button in a sensible way wasn't easy. I tried tabs (like in trading) but they took too much space, especially valuable on smaller screens. I ended up with this:

image
image

The button from collection to missions on mobile was especially hard, I ended up with this:

image

Which sadly adds another row 😕 This is an alternative, doesn't add a row, but is a little bit awkward:

image

I was also considering adding another navbar tab, but it's also not great. Missions aren't that important to consume this precious space in my opinion. Any ideas on how to better handle going to missions are really welcome!

### DIFF
--- a/frontend/public/locales/en-US/common.json
+++ b/frontend/public/locales/en-US/common.json
@@ -1,0 +1,4 @@
+{
+  "noAccount": "Account not found",
+  "loading": "Loading"
+}
--- a/frontend/public/locales/en-US/pages/collection.json
+++ b/frontend/public/locales/en-US/pages/collection.json
@@ -1,4 +1,6 @@
 {
+  "goToCollection": "Go to collection",
+  "goToMissions": "Go to missions",
   "publicCollectionTitle": "You are viewing the public collection of \"{{username}}\".",
   "publicCollectionDescription": "You are currently viewing a public collection. Click any of the navigation items to return to your own collection.",
   "showPossibleTrades": "Show possible trades",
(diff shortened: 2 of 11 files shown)


marcelpanse:
Since missions aren't used that much and not very important, can we hide this button and only show the button in the "all filters" popup instead? 
Then the mobile version also needs an "all filters" button which should only show the available filters (expansion, owned/missing), which would be consistent. And a button to go back to the collection. 

Clicking on collection in the navbar should also always bring your back to the normal collection. This should be the case assuming you made a new route for the missions (like /collections/missions) - I didn't look at the code yet nor tested anything.

greg19:
Having a separate route for missions makes so much more sense than what I tried initially. Since there are only two relevant filters for the missions, I don't think the "all filters" button is needed, they currently fit in two lines:

image

PR #642, opened 2025-09-16
Show list of people actively trading

### LINKED ISSUE #609: Show list of people actively trading
The idea is to make another tab in the trading section that shows a list of possible trading partners. To show which cards you are looking for I'm thinking to have a favorites option (similar as in the game). You can favorite lets say up to 5 cards (note; we don't need 20 like in the game, because we already can see all the cards you need for trading, this is just to showcase your top 5 in the list). 
The 5 favorite cards we can store in a simple array on the account table and another column to store the timestamp when you last updated your favorite list maybe. 

When you turn on active trading and add at least add one favorite, we'll show you on the trading list page (ordered by the timestamp). 

To prevent people spamming the trading list, there should be a cooldown. We can do this by only updating the timestamp if the timestamp is older than 1 day. 

The favorites should also be tradable cards, ignoring the min/max filters. This has the added benefit that if you are looking for one specific card for a deck even though you already have it it will show in your tradables. 

On the trading list itself, we can highlight the matches where you have duplicate copies of the favorite card requested.

This is just my initial idea, let me know if anyone has ideas/remarks.

marcelpanse:
closes #609

- [x] add the cards table
- [x] update the query to join with the cards table
- [x] create UI with list of users and links to their trading pages

### DIFF
--- a/frontend/public/locales/en-US/trade-matches.json
+++ b/frontend/public/locales/en-US/trade-matches.json
@@ -3,7 +3,7 @@
   "settingsTitle": "Trade settings",
   "openSettings": "Open trade settings",
   "openTradeWith": "View matches",
-  "viewTradePartner": "View {{matched_cards_amount}} matches",
+  "viewTradePartner": "View matches",
   "tradingWith": "Trading with",
   "noTradeOffers": "No trades yet, head to the community to find someone to trade with!",
   "noTradePartners": "We couldn't find anyone for you to trade with.",
--- a/frontend/src/pages/trade/TradeWith.tsx
+++ b/frontend/src/pages/trade/TradeWith.tsx
@@ -1,9 +1,6 @@
-import { CircleHelp } from 'lucide-react'
 import { useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useParams } from 'react-router'
-import { Tooltip } from 'react-tooltip'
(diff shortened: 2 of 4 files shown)


marcelpanse:
This fetches the list of potential partners. Note, it doesn't work that well yet if you max number of cards is set to 1 since it won't search for cards you don't have any copies of yet - we need the cards table for that. Next steps:
- add the cards table
- update the query to join with the cards table
- create UI with list of users and links to their trading pages

greg19:
If I understand correctly, the current implementation counts the number of cards you can give + the number of cards you can receive. For example, this is my number one with 254 "trade matches":
image
And this one is supposed to have 169 matches:
image
Both numbers are clearly too high, I think you incorrectly count the rarities that cannot be traded.


With this metric, people with almost full collection will be the ones almost exclusively getting pushed to the top (as seen above). That's not very good for the site:
- there might be a possibility get overwhelmed by the number of requests they get
- other people will get much less requests
- it's just not a good recommendation on it's own, I'd rather have 20 cards to choose from, but being able to make 15 trades total, that having 50 cards to choose from but being able to make 2 trades total

I think a better metric would be the number of actual trades you can make. I guess it should also more uniformly distribute the rankings. Did you see any issues with the last query I proposed in #609? I haven't been able to test it, but it should count exactly that. Also, is there a way I can run deno with local changes, to test the query?

PR #664, opened 2025-09-28
Refactor filters

greg19:
I was trying to tackle [this](https://community.tcgpocketcollectiontracker.com/t/ability-to-show-trade-looking-for-cards-externally/6933), but editing existing filters was really cumbersome, with each of them having it's own `tsx` and translation files. The filters were often very similar (for example expansion/sort-by/number are all just dropdowns, or type/rarity are select many lists) with a lot of code duplicated. Also each filter was styled independently, with some having background and some not.

In this PR, I extracted the duplicated code to a reusable filter types that you can use with any filter values you want. Also made their style more cohesive. And merged the translation files for less fragmentation (many had just a single word in them).

image
image
image
image
image
image
image
image


marcelpanse:
Nice, look good!

Regarding the forum thread you mentioned. I created a separate issue here https://github.com/marcelpanse/tcg-pocket-collection-tracker/issues/666 with some thoughts

PR #672, opened 2025-09-30
feat: add and adapt missing french translations

nicolasfeyer:
@afu-dev Could you proofread this as a second French-speaking set of eyes? :)


afu-dev:
And thank you for your contribution!

Issue #673, opened 2025-09-30
Support for new deluxe set


marcelpanse:
As far as I can tell from opening some packs a couple of major issues are:
- only 4 cards in a pack instead of 5
- 4th card is always an EX
- packs contain old cards, however they seem to be treated as new cards.
- it you get an old card, the old card in the dex will still show as a question mark, so it seems you don't have it
- until you click on it and it shows you have one (but in the new set)
- the new set also shows question marks for cards you already have in other sets 
- BUT I did get all the missions immediately since I have the cards in old sets

So I don't even know if the probability calculation of a new card should be purely based on the new set or a combination of the old and new sets.

Kind of a mess.

Tasks:

- [x] add new set & cards
- [x] implement pull rates for 4 card sets
- [x] implement pull rates for "deluxe packs" --> '4th card is always an EX or better'
- [x] verify is probabilities for previous packs didn't change
- [x] verify probabilities of new deluxe pack
- [x] add set selection before scanning cards
- [x] test scanner with old cards
- [x] test deck building mode
- [x] foil cards in scanner are not working, implement solution
- [x] test scanner with foil cards
- [x] implement new database structure
- [x] refactor scraper to create the internal_ids properly and link them
- [x] adjust code to work with new structure
- [x] write migration script to update the db to the new structure
- [x] make sure local storage gets invalidated by using a new cache key
- [x] deploy & run migration

marcelpanse:
Another issue to take into account is that the old and new cards are identical which will be problematic for the scanner, since which of the two will be matched and incremented?

drdenjef:
Can the number of cards in an image be used? If it is four, it is from the deluxe set.
I know this will not work with cards from WP's. For that I do not have an idea right now that does not require user input.

marcelpanse:
> if I didn't already know I pulled three from one set and three from another, how would I find out?
it seems in the game you can't.. 

> So, clearly, the game has internal differentiating counters that it doesn't show me.
That's the main problem here. In the game they are treated as separate cards. But if you click on one of the versions, it shows as one card. And in the card counts it's all mixed up as well.

JakeAlper:
In that case, it seems the *only* way to keep track of how many "connected" cards are pulled from each expansion is by manually specifying which pack was opened when receiving it...

The amount of each of these cards does actually affect the dex missions, too, so... that's a mess. The dex missions feature might be inaccurate if the user doesn't track every single card and which pack they opened it from.

[...]

PR #674, opened 2025-09-30
deluxe set support

marcelpanse:
See: #673


frontend/src/types/index.ts
```diff
@@ -62,6 +62,14 @@ export interface TradePartners {
   matched_cards_amount: number
 }
 
+export interface PackStructure {
+  cardsPerPack: 4 | 5
+  guaranteedRarity?: {
+    position: number
+    rarity: Rarity
+  }
```
greg19, line 70: I think you don't need this, setting the correct probability distribution should be enough. Is it even used anywhere?

frontend/src/lib/CardsDB.ts
```diff
@@ -420,6 +443,105 @@ const probabilityPerRarityBaby: Record<Rarity, number> = {
...
+  '◊◊◊◊': 100, // GUARANTEED
+  '☆': 0,
+  '☆☆': 0,
+  '☆☆☆': 0,
+  '✵': 0,
+  '✵✵': 0,
+  'Crown Rare': 0,
+  P: 0,
+  '': 0,
+}
```
greg19, line 509: Maybe use an array to avoid the awkward 1234 variables.

PR #682, opened 2025-10-05
#673 refactor card structure

marcelpanse:
Relates to #673

### DIFF
--- a/frontend/assets/cards/A1.json
+++ b/frontend/assets/cards/A1.json
@@ -26,7 +26,7 @@
     "pack": "mewtwopack",
     "alternate_versions": ["A1-1", "A1-227", "P-A-23", "A3-210", "A4b-1", "A4b-2"],
     "artist": "Narumi Sato",
-    "internal_id": 262273,
+    "internal_id": 65600,
     "linked": false
   },
   {
@@ -56,7 +56,7 @@
     "pack": "mewtwopack",
     "alternate_versions": ["A1-2", "A3-211", "A4b-3", "A4b-4"],
     "artist": "Kurata So",
-    "internal_id": 262402,
+    "internal_id": 65665,
     "linked": false
   },
   {
@@ -86,7 +86,7 @@
     "pack": "mewtwopack",
     "alternate_versions": ["A1-3", "P-A-18", "A3-212"],
     "artist": "Ryota Murayama",
-    "internal_id": 262531,
+    "internal_id": 65730,
     "linked": false
   },
   {
@@ -122,7 +122,7 @@
     "pack": "mewtwopack",
     "alternate_versions": ["A1-4", "A1-251", "A3-230", "A4b-5"],
     "artist": "PLANETA CG Works",
-    "internal_id": 262660,
+    "internal_id": 65795,
     "linked": false
   },
   {
@@ -152,7 +152,7 @@
     "pack": "pikachupack",
     "alternate_versions": ["A1-5"],
     "artist": "Miki Tanaka",
-    "internal_id": 262785,
+    "internal_id": 65856,
     "linked": false
   },
   {
@@ -182,7 +182,7 @@
     "pack": "pikachupack",
     "alternate_versions": ["A1-6"],
     "artist": "Yuka Morii",
-    "internal_id": 262913,
+    "internal_id": 65920,
     "linked": false
   },
   {
@@ -216,7 +216,7 @@
     "pack": "pikachupack",
     "alternate_versions": ["A1-7", "P-A-13"],
     "artist": "Shin Nagasawa",
-    "internal_id": 263043,
+    "internal_id": 65986,
     "linked": false
   },
   {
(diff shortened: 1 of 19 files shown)


scripts/scraper.ts
```diff
@@ -253,22 +247,50 @@ async function extractCardInfo($: CheerioAPI, cardUrl: string, expansion: string
...
+  const ex = name.includes(' ex')
 
   // Check if card is a baby pokemon (Not currently specified exactly on Limitless TCG page)
   const baby = weakness === 'none' && hp === '30' && energy !== 'Dragon'
 
   const { setDetails: set_details, pack } = extractSetAndPackInfo($)
 
   const alternate_versions: string[] = []
+  let linked = false
+  let internal_id = ''
```
greg19, line 259: What do you think of my recommendation of using integers for internal ids? They should be much more performant.
- String cause additional heap allocations
- Strings are much slower to compare with `===` than integers
- If you want to have some mapping from the ids to anything, using either `Record<string, T>` or `Map<string, T>`, both of these will actually be hash lookups that are much slower than array access (they hash your string into an integer and use this integer to access underlying array, this hashing is a considerable overhead, and additional collisions can make it even slower), and also take more memory (they allocate an underlying array that is much bigger than the number of elements to deal with the possible collisions).

I don't really know, but I expect these index operations will be very common. So I wouldn't be surprised if using `string` here instead of an integer would make the whole app like 10-15% slower compared to using integers.

How I would generate the ids:
While the scrapper is running, I'd keep an array `internal_cards` that would hold all the cards encountered so far. Then, in `extractCardInfo`, I'd loop through the array and look for any matches (using `internal_cards.findIndex`). If there is a match, use it's index as the `internal_id`. If there isn't any match, append a new internal card at the end. This would probably mean we would need to limit the `async` concurrency in the `scrapeCards` though, but I think that's fine.

greg19: Okay, I knew engines optimized small records that appear often but didn't know that thousand field ones that you have a single object of. I think we will need to remember to use `Record` over `Map` to benefit from these optimizations (correct me if I'm wrong).

If these are hashes of the `A4b-1` kind of strings, why not use them directly?

marcelpanse: > I think we will need to remember to use Record over Map to benefit from these optimizations

Yes, correct. Map is optimized for loops, so insert is slower, but iteration is faster compared to js objects but lookups are slower. See: https://share.zight.com/2NuDZAmm. So using the standard js objects is best for our case if we want to optimize for lookups. 

To improve performance I could also write the json objects directly as objects in the json like:
```
{
  "A1-1": {...},
  "A1-2": {...}
}
```

... so we don't have to make these dicts `const allCardsDict: Map<string, Card> = new Map(allCards.map((card) => [card.card_id, card]))` anymore and can do directly `cards['A1-1']` to get a card.

However, you can't loop them directly, so you have to do:
```
for (const value of Object.values(test)) {
  console.log(value);
}
```

> If these are hashes of the A4b-1 kind of strings, why not use them directly?

I was also contemplating that. I just found a separate table with internal ids that are existing card id's a bit confusing if you look at it. But in practice it doesn't really matter. I don't mind either, do you prefer to use "A1-1" to use as the internal_id instead?

greg19: The number construction you suggest is actually pretty clever, I like it 🙂

I'd work in bit level though, it will be easier, you have bit shifts/masks and it's blazing fast 😉

Assiming we use 32 bit integer as the ID, we might simply allocate 16 bits to expansion and 16 bits to within-pack number. Realistically, 10 and 12 bits respectively is probably safe so we might even use the 10 extra bits for something else, but I don't know if there would be something useful (pack? pokemon/trainer? rarity?).

We might just order expansions by the time their are released to not deal with the prefixes/suffixes, and be safe if they suddenly release something strange (though this can suffer from the reordering issue you are talking about, but to much lesser scale). Then the internal id would be just `(expansion_number << 16) | card_number_within_pack` (if using 16 bits per field). If you want to encode all the parts separately, you might want to allocate more bits to the expansion just to be safe, and maybe leave some spare for future use in case they release something strange.

greg19: I realised that if we encoded rarity in the ID, we would be able to filter the tradable cards in the SQL, without the need of any external `tradable_cards` table 🤯

greg19: Treat all expansions equally:
```
Promo -> 0
A1 -> 1
A1a -> 2
A1b -> 3
A2 -> 4
A2a -> 5
A2b -> 6
A3 -> 7
A3a -> 8
A4 -> 9
A4a -> 10
A4b -> 11
```
And each new expansion would get a new number, regardless of it's code. This way we are safe if they suddenly release something strange like `Super1a` or whatever weird they might come up with. We would just keep giving it the next free number.

In code, we would keep a simple array of all the expansions and use `chronological_expansions.indexOf(expansion)` as this number. (Promos should be at the beginning to not mess up when adding new expansions at the end)

Issue #680, opened 2025-10-03
Deluxe Pack Bug - 100% Chance Deckbuilding


Gigglestomp:
## Describe the bug

Shows 100% chance of a new card. But I have deck building turned on.

## To Reproduce
Toggle deck building. Still shows 100%.

## Expected behavior

No way its 100% I have all the non foil or non art cards at 4 diamond and below.

![Image](https://github.com/user-attachments/assets/3b9de58a-4789-4123-ad6c-bc598a25d64b)

## Desktop (please complete the following information):
 - OS: [e.g. Windows, ]

marcelpanse:
I'm guessing the code for the deckbuilding mode does not take into account that there are now similar cards in the same set of the same rarity (the foil vs non-foil).

Gigglestomp:
Closed in error when replying. Can't see how to reopen.

chronofreak:
No, I have "Number of cards wanted" at 2 with deck-building mode ON, since I'd like any 2 of each card regardless of version/rarity. I assumed setting qty to 2 with deck-building on would count any version of a card..? Example, having 1 single-diamond Bulbasaur and 1 full-art Bulbasaur would result in my collection having a total of 2 Bulbasaur as far as deck-building mode is concerned.

I also get 1.5% with the number set to 1 and just Solgaleo missing from my collection.

Using the filters+deckbuilding mode, I've verified that Solgaleo is the only card that I have less than 2 of. Maybe I'm doing something wrong or have incorrect expectations of how the deck-building mode works?

![Image](https://github.com/user-attachments/assets/8a364e89-5fa0-488e-8f45-0587b5fcf0d5)

![Image](https://github.com/user-attachments/assets/e4e1ee74-46d4-4f34-bf6f-026784733ef8)

![Image](https://github.com/user-attachments/assets/a74b0fca-71e4-49d8-b101-cad8043260f3)

PR #703, opened 2025-10-17
Fix deckbuilding

### LINKED ISSUE #698: Deck building amount doesn't account for internal IDs
Image

The migration to internal cards didn't migrate "alternate versions" (it still uses the old card_ids). Because of this all the `card.alternate_versions.reduce` pull out the same card multiple times, resulting in an incorrect total amount.

We may consider some helper `deckbuilding: Card[] => Card[]` function to calculate the deck building amounts to not hunt for every `card.alternate_versions.reduce` in the codebase in the future.

greg19:
Fixes #698.
~Maybe also #680.~
Fixes Androsa issue from [forum](unity.tcgpocketcollectiontracker.com/t/cards-present-in-the-deluxe-pack-count-twice-for-deck-building-mode/7042/11).

- Added more robust type checking to scraper
  - Fixed a bug where fossils had `energy = '40 hp'`
  - Fixed a bug where shiny rarities were sometimes counted in deckbulding mode – in some places it relied on the `fullart` field, which was incorrectly set for them. Switched to using `basicRarities` everywhere and removed the `fullart` field entirely (it can be reproduced again with `[...].included(card.rarity)` if needed, but doing it explicitly will avoid bugs like this).
  - Hidden some irrelevant fields from `CardDetail` for trainers (like retreat, weakness, hp)
- Made scraper use `internal_id`s for `alternate_versions`
- Made frontend use `intenal_id`s when calculating alternate versions. I didn't check all the code paths for deckbuilding, so there might still be some hidden issues.

Note that the `getCardByInternalId` gives you the card with `card_id` from the latest expansion (as seen on the screenshot below) and I just used it as-is. In the future, when we change the data model to a normalized one (ditching card_ids all together), we should change how the `CardLine` displays the cards (IMO the most sensible one: ◊◊◊◊ Leafeon (_A2a, A4b_) x1). But until then we can just keep using the latest expansion card_id when we need to show one.

image
image

### DIFF
--- a/.github/workflows/verify-hashes.yml
+++ b/.github/workflows/verify-hashes.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup Node 22
-        uses: actions/setup-node@4
+        uses: actions/setup-node@v6
         with:
           node-version: 22
           check-latest: true


marcelpanse:
> @marcelpanse can the `Card.linked` attribute be removed now? Or you want to do some additional checks?

That can be removed, it was just to check if the scraper linked them together correctly.

greg19:
If you show both, the amounts won't visually add up to the number on the bottom, which is IMO more confusing

greg19:
> You do have both card_ids right?

No, the `alternate_versions` changed to `internal_id`s in this PR (so the list got physically shorter). And the `getCardByInternalId` is currently defined as:
```ts
const allCardsByInternalId: Map<number, Card> = new Map(allCards.map((card) => [card.internal_id, card]))
```
So it gives you the latest card when there are duplicates. We can change it to
```ts
const allCardsByInternalId: Record<number, Card[]> = Object.groupBy(allCards, (card) => card.internal_id)
```
But we would need to add extra logic in every place where `getCardByInternalId` is called to handle multiple versions everywhere (and in many places this might not even make sense). I think at the end of the day we are aiming to handle cards by their `internal_id` (and allow a single card to be in multiple expansions), so this would be rather counterproductive.

Issue #666, opened 2025-09-28
Create preset filters and merge trade-cards/collection pages


marcelpanse:
I was thinking that the trading 'looking for' and 'for trade' should both be in a "preset filter" dropdown that you can select in the all filters panel. These should be deeplinkable like https://localhost:5173/#/collection?filters=for_trade. We can also think of other useful presets for filters. 

Then we can completely remove the Cards tab on the trading page, which in my opinion should simply be part of the collection page anyway. That would simplify the code / make it easier to maintain and it makes it more consistent in the UI.

greg19:
I don't know what you mean by "preset filter", but I had similar vision: move the filter to collection, and remove "Trade/Cards". The "copy trades to clipboard" will also need to be moved, and placed somewhere. There is a "share" button already (that for some reason opens "Edit profile" currently), so it might be a good place.

With the link, I was also thinking about that. My idea was to serialize all `filter` fields into GET parameters.

PR #704, opened 2025-10-20
Deeplinkable collection

greg19:
Per #666

- [x] Moved filtering logic outside `FiltersPanel` component
- [x] Make the URL the source of filter params. I also considered using `useState` and pushing the state to the URL, but always deserializing from the URL allows to act when the user modifies the URL by hand
- [x] Make the browser URL the sharable link for people with public collections


greg19:
I used native browser share feature `navigator.share`. It's nice on mobile, you can share the link in a few click on other apps, like WhatsApp. But it isn't super pretty on desktop. Do you think it's worth it, or stick with clipboard (it already falls back to clipboard if `navigator.share` is not supported on the device)?

We can also try to detect user agent, but it's a hack that won't be 100% accurate and will probably sooner or later break in some weird way, so it will be harder to maintain (analogously to the recurring problems with the fancy 3d view of the cards).

### Mobile
image

### Desktop
image

Issue #705, opened 2025-10-20
Carddex status not working for public collections


greg19:
<img width="945" height="794" alt="Image" src="https://github.com/user-attachments/assets/668f32d2-5954-41c8-8419-110c0c07be4e" />

<img width="932" height="840" alt="Image" src="https://github.com/user-attachments/assets/263ccc2d-d94b-49db-9dc6-4989d70f7f77" />

greg19:
There is also [column level security](https://supabase.com/docs/guides/database/postgres/column-level-security), maybe it might be helpful (although supabase warns that it is a advanced feature).

marcelpanse:
That doesn't work if I understand correctly, it applies a list of columns you can read to a certain role (like authenticated). In our case your role is authenticated and you want to be to read the email column, but only on your own rows, then on other rows, you want to read certain columns except the email column. So it's more like RLS, but with a limited set of roles.

I think I found a different solution where I add the needed columns in the view itself. I defined a new view that adds the collected card_ids to the view output so we can use that to solve this problem. Like this: 

<img width="354" height="167" alt="Image" src="https://github.com/user-attachments/assets/0939eec3-eb38-4216-9aa5-d4236f58f4ee" />

I can do something similar for the public_accounts / trade_rarity_settings. Another option would be to not put it in a separate table, but just add a json column on the account object with the trade settings object in it. That would make the DX simpler, but will likely make the performance of the trade matches sql worse.

PR #730, opened 2025-11-06
Trading rarity settings

### LINKED ISSUE #662: Trade matches algorithm could be improved
My current best match looks like this:

Image

Where I have a lot of ◊ and ◊◊ to give, but not too much to receive, and the other way around for ◊◊◊ and ◊◊◊◊. I believe the current algorithm calculates the number of trades you can make as the minimum of the number of cards you can give and the number of cards you can receive. As seen on my example, the true number of trades I can make with the person is much lower. As I raised in the code review, to get the correct number, you need to take the minimum in each rarity, and then sum it.

There were voices in the forum [link](https://community.tcgpocketcollectiontracker.com/t/new-trade-feature/6689/16) that people see matches they cannot trade with at all. I believe the reason might be that the disproportion is 0-x in one rarity and x-0 in other, there are no actual possible trades, but the algorithm incorrectly counts it as a lot of possible trades.

### LINKED ISSUE #726: Add rarity filter for trade matches
In the trade settings there should be a rarity select where you can select multiple rarities. If you do, trade partners/matches should be only matched on the selected rarities. 

I'm mostly interested in trading full-art cards and don't want all the clutter of the diamond cards of the current set (since i'm still opening packs). Also I don't card about the A4b foil cards either, so I could just filter to match only on the full-art rarities.

greg19:
Closes #726. Closes #662.

This adjusts the frontend logic, the trade matches SQL will need to be changed later. @marcelpanse you can create the table (`trade_rarity_settings(email, rarity, to_collect, to_keep)`) and modify the settings UI now. For public accounts, we will have the same problem with joins as in #705, so it might be worth to check out the access to the underlying `account` (as opposed to the `public_account`) via conditional RLS based on the `account.public_account` column as in my comment.

### DIFF
--- a/frontend/public/locales/en-US/trade-matches.json
+++ b/frontend/public/locales/en-US/trade-matches.json
@@ -55,5 +55,6 @@
   "raritySettings": "Rarity Settings",
   "rarity": "Rarity",
   "toCollect": "To collect",
-  "toKeep": "To keep"
+  "toKeep": "To keep",
+  "longOperation": "This might take a while."
 }
--- a/frontend/src/pages/trade/TradeMatches.tsx
+++ b/frontend/src/pages/trade/TradeMatches.tsx
@@ -8,15 +8,21 @@ function TradeMatches() {
   const { data: tradingPartners, isLoading, isError } = useTradingPartners()
 
   if (isLoading) {
-    return <p>{t('common:loading')}</p>
+    return (
+      <p className="text-xl text-center py-8">
+        {t('common:loading')}
+        <br />
+        {t('longOperation')}
+      </p>
+    )
   }
 
   if (isError) {
-    return <p>{t('common:error')}</p>
+    return <p className="text-xl text-center py-8">{t('common:error')}</p>
   }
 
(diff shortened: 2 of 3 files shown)


marcelpanse:
@greg19 I'm not sure about the changes in the merge conflict, could you fix that for me?

marcelpanse:
Localhost 2025-11-16 at 7 45 36 PM

greg19:
I can look into the SQL. Do we have some testing environment where we can test the function before the PR merged? Or how does that work?

marcelpanse:
I don't have a testing env unfortunately. Only the Supabase production account. Maybe I can make a readonly user for the postgresdb then you can connect directly to that db and run your sql to test it. 
Once ready, i'll deploy the supabase function to prod. Would that work?

marcelpanse:
Regarding performance optimization, GPT has some good suggestions to improve performance. --> https://chatgpt.com/share/691f776a-9144-8008-b289-f0b6f661ffb0

marcelpanse:
updated the indexes & queries. seems good to go live

Issue #711, opened 2025-10-23
Overview card count uses `card_id`s instead of `internal_id`s


greg19:
## Describe the bug

There are 80 ◊◊◊◊ cards in the game, but overview (without deckbuilding) shows 155 (which looks like 80 from A1-A4a and 75 duplicates from A4b).

## To Reproduce
1. Go to Overview
2. Apply a ◊◊◊◊ filter
3. See 155 total cards

## Expected behavior
There should be 80 total cards.

## Screenshots

![Image](https://github.com/user-attachments/assets/980408e1-0a7a-422e-9cdf-2b530c8de6ed)

PR #734, opened 2025-11-09
Deck builder

greg19:
A new page where you can construct your own decks. For now, they are not synced to the DB, to not commit to any schema too soon. The only way to save the deck now is to save the link to it (and you can also use the link to share it to other people!). My idea is to first get some feedback from the users, and then decide on the data shape. But any feedback before merging is also valuable :).

I changed the `FiltersPanel` component to be more flexible and tweaked it in the collection page, feedback here is encouraged :). This filtering logic will also ready to be easily plugged into the overview page in the future (which will probably solve #711).

I didn't re-use the `CardsTable` component because it wasn't flexible enough (I didn't want to include the card id, collection status and card count to have a cleaner UI). What's interesting, the deck builder page opens instantly, while the collection page takes like 3-4 seconds to load (on prod) which suggests to me that it could be improved. I was also able to normally include a scroll bar which got manually disabled for some unknown to me reason in collection page (which is really annoying). And I didn't face any problems with updating the table on filter change (there is a comment that says it is in some way "hard" in the collection, and introduces a lot of unintuitive workarounds for it).

I'll probably separate the `FiltersPanel` and `Collection` changes into another PR where I might also do some investigation on the above possible improvements.

I also partially addressed #735 (but no changes to the scraper yet).

### Deck builder page

image
image

### Deck list page

Screen Shot 2025-11-09 at 01 12 25
Screen Shot 2025-11-09 at 01 13 08

### DIFF
--- a/frontend/src/pages/decks/DeckBuilder.tsx
+++ b/frontend/src/pages/decks/DeckBuilder.tsx
@@ -13,7 +13,7 @@ import { Sheet, SheetContent, SheetHeader, SheetTitle } from '@/components/ui/sh
 import { showCardType } from '@/components/utils'
 import { allCards, getCardByInternalId } from '@/lib/CardsDB'
 import { type Filters, type FiltersAll, getFilteredCards } from '@/lib/filters'
-import { useCollection } from '@/services/collection/useCollection'
+import { useCollection, useSelectedCard } from '@/services/collection/useCollection'
 import { type Energy, energies } from '@/types'
 import { serializeDeckToUrl } from './utils'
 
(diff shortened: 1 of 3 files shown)


greg19:
Things left to do:
- Implement mobile view
- Add editable energy used
- Add editable deck name

marcelpanse:
It looks cool! Is this ready for review/merge? 

Couple of notes:
- there is an error in the console on the page: "In HTML, <button> cannot be a descendant of <button>."
- when clicking on the card I expect the sidepanel to open with card details since it also works like that in the rest of the site. There should be a plus button next to the name below the card to add the card to the deck. 
- I don't understand what the what the element type multi-select does on the rights side of the screen

PR #741, opened 2025-11-13
fixes #705 - carddex status not working for public collections

### LINKED ISSUE #705: Carddex status not working for public collections

marcelpanse:


frontend/src/types/index.ts
```diff
@@ -34,6 +34,7 @@ export interface CollectionRow {
   internal_id: number
 
   collection: string[] // array of cardIds
+  collected_card_ids?: string[]
```
greg19: Shouldn't this be one thing?

PR #744, opened 2025-11-15
Remove `cursor-pointer` from a button

greg19:


greg19:
I see, so we should remove this one added manually

PR #750, opened 2025-11-19
Fix P-B internal ids

### LINKED ISSUE #749: Collection: P-B-5 Pidgey linked to B1-180 Pidgey, and counts incorrectly.
## Describe the bug

There is an issue with how card P-B-5 Pidgey is counted in the "Collection" tab.

The count for P-B-5 Pidgey seems to be connected to the count for B1-180 Pidgey so that when you increase or decrease the count for 1 card, the count changes in sync for the 2nd card as well, like the Deluxe Ex cards do with their matching card.

## To Reproduce
1. Filter to "Pidgey
2. increase or decrease the count on either P-B-5 Pidgey or B1-180 Pidgey
3. Observe that the count for the other card changes with it.

greg19:
Fixes #749 for now, but we will need to rethink it. Community posts: [one](https://community.tcgpocketcollectiontracker.com/t/wrong-card-update/7165) and [two](https://community.tcgpocketcollectiontracker.com/t/pidgey-card-count-error/7160/5).

### DIFF
--- a/frontend/assets/cards.json
+++ b/frontend/assets/cards.json
@@ -20738,7 +20738,7 @@
     "pack": "arceuspack",
     "alternate_versions": [
       263744,
-      12648516
+      12648528
     ],
     "artist": "MAHOU",
     "internal_id": 263744
@@ -68506,7 +68506,7 @@
     "pack": "megablazikenpack",
     "alternate_versions": [
       788544,
-      12648898
+      12648912
     ],
     "artist": "GOSSAN",
     "internal_id": 788544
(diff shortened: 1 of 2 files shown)


marcelpanse:
I assume this will break as soon as we run the scraper?

greg19:
No, I added an override for the whole P-B:

```js
const rarityOverrides: Partial<Record<ExpansionId, { rarity: Rarity; start: number; end: number }[]>> = {
  // ...
  'P-B': [{ rarity: 'P', start: 0, end: 999 }],
}
```

PR #776, opened 2025-12-04
Improve scan UI

greg19:
- You no longer need to specify the expansion before scanning
- If a card appears in multiple expansions, you will be able to select which one to use (defaults to the latest), see the screenshot
- Changed the keys in `hashes.json` from `card_id` to `internal_id`
- You now see your current amount owned during scanning, and how it will change. It might be helpful in catching any desynchronization since they added the amounts to the pack opening screen (so you can compare the screenshot amount to what the site tells you).
- You can now independently select if you want increment or decrement a card. Useful for trading.

image


marcelpanse:
That's a nice improvement, thanks!

PR #780, opened 2025-12-05
Add page titles

greg19:
Adds more descriptive page titles to: trade with, friend collection, deck builder. So you can use your browser history to navigate to them.

For example when you receive a trade from a friend (that you didn't agree on on the site) and want to quickly check what cards they need (type their username into the browser and you will get a history entry).


greg19:
React 19 does this :) [link](https://react.dev/blog/2024/12/05/react-19#support-for-metadata-tags)

marcelpanse:
Ow, I didn't they added support for that!

PR #782, opened 2025-12-06
Find card

greg19:
Allows you to search for a specific card to trade.

image


greg19:
@marcelpanse can you help me setting the data fetching here? The request should me made when the users clicks the button, and I don't know how to set up tanstack query this way.

greg19:
> > I was not sure what you intended that should happen when you search on a specific card (like bulbasaur). Currently, it finds a good match, but when you click on the "view matches" button, you'll go to the trading page, but the bulbasaur card I was searching for isn't visible there since it doesn't match my trading settings.
> 
> Probably we can pre-select the card on the TradeWith page and it will be fine (and show you only the same rarity cards to offer on your side).

Done, should be good to go.

The last commit is larger because it fixes a bug where the `selectedCard` was used even when looking for all matches (if you selected some card, but clicked "search all" it would still search only for the single card).

PR #801, opened 2025-12-16
feat: added missing translations and did the french version

afu-dev:
- Added missing "socials" translation (share on, copy trading link, copy collection link).
- Added missing "You can now manage your trading cards on your collection page" translation.
- Translated to french all string from the trade system.
- Translated to french new trading filters.
- Translated to french the "trade matches" page.
- Translated Pokémon names in the trade message when copiing it to clipboard.

Note to discuss:

https://github.com/marcelpanse/tcg-pocket-collection-tracker/pull/801/changes#diff-a99fc3add5c830820ff40f66802ecabc5da7594dab4ce5e1519519ad0dbdb55fL62

https://github.com/marcelpanse/tcg-pocket-collection-tracker/pull/801/changes#diff-b42ce21162993e6f1cd8a08a5e5aa9ac638cf72ceaad3345a3c8543c6aed7d61L28-L29

In the Trade > Matches page, I had to translate to french the Pkmn names. But in this page, we are using the `getCardByInternalId`, which can return undefined. I've updated the `getCardNameByLang` method to allow undefined and return an empty string in that case. But I'm not 100% happy with the solution. If I can do anything better, please don't hesitate to tell me 😄

### DIFF
--- a/frontend/src/lib/utils.ts
+++ b/frontend/src/lib/utils.ts
@@ -25,8 +25,8 @@ export function formatFriendId(friendId: string): string {
   return `${friendId.slice(0, 4)}-${friendId.slice(4, 8)}-${friendId.slice(8, 12)}-${friendId.slice(12, 16)}`
 }
 
-export function getCardNameByLang(card: Card | undefined, lang: string): string {
-  if (card === undefined || card.name === undefined || card.name === null) {
+export function getCardNameByLang(card: Card, lang: string): string {
+  if (card.name === undefined || card.name === null) {
     return ''
   }
 


greg19:
> In the Trade > Matches page, I had to translate to french the Pkmn names. But in this page, we are using the `getCardByInternalId`, which can return undefined. I've updated the `getCardNameByLang` method to allow undefined and return an empty string in that case. But I'm not 100% happy with the solution. If I can do anything better, please don't hesitate to tell me 😄

You can instead extract the card to a variable:
```js
const card = useMemo(() => getCardByInternalId(selectedCard), [selectedCard])
```
And then have the ternary match on this object:
```js
{card ? t('search.byCard', { cardName: getCardNameByLang(card), i18n.language) }) : t('search.selectCard')}
```
A nice thing to have then is a sanity check (that would never trigger, unless there is something wrong with the card database – in which case a clear signal helps debugging):
```js
if (selectedCard && !card) {
    return /* error alert */
}
```

frontend/src/lib/utils.ts
greg19: Now you can change this back

afu-dev: nvm! Because of the check on card later, I can only have a Card type when calling the getCardNameByLang. Fixed!

greg19 (approved):

PR #802, opened 2025-12-16
feat: added the missing and owned filter option back

afu-dev:
Please check the discussion on the community forum before: https://community.tcgpocketcollectiontracker.com/t/missing-owned-changed-to-wanted-for-trade/7264

### DIFF
--- a/frontend/public/locales/de-DE/filters.json
+++ b/frontend/public/locales/de-DE/filters.json
@@ -20,5 +20,7 @@
     "all": "All",
     "wanted": "Wanted",
     "extra": "For trade"
-  }
+  },
+  "carddex": "Card Dex",
+  "trading": "Trading"
 }
--- a/frontend/public/locales/en-US/filters.json
+++ b/frontend/public/locales/en-US/filters.json
@@ -1,5 +1,7 @@
 {
   "rarity": "Rarity",
+  "carddex": "Card Dex",
+  "trading": "Trading",
   "f-ownership": {
     "all": "All",
     "missing": "Missing",
--- a/frontend/public/locales/es-ES/filters.json
+++ b/frontend/public/locales/es-ES/filters.json
@@ -20,5 +20,7 @@
     "all": "All",
     "missing": "Missing",
     "registered": "Registered"
-  }
+  },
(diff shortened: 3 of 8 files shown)


afu-dev:
> `maxAmount` = 0 is not the same as Missing cards. A card is missing if you don't have it in your dex, but you can still have it in your dex but with a 0 amount.

(Totally didn't know we could have a owned card with zero amount)

And after testing, we currently have bugs in production regarding 0 amount-owned-cards:

image
(changing the amount of Mega Pinsir crash the app)

greg19:
> (This is just to challenge the idea)
> 
> A "missing" filter seems way more obvious than "maximum numbers of card is zero". The first is "linguisticly"/"semantically" correct, the second is "mathematicaly" correct. A missing filter doesn't require you to think about what it does, it just shows missing cards. The "max at zero" forces the user to think "ok if I have the _maximum number_ of my cards at zero, this means I don't have it, so I'm missing it". At least in my brain, there is in extra step in the thinking.
> 
> Btw, the interface is for the general public, not specifically developers like us. And I think the reflection is a developer bias.

Yea, I know. I'm trying to counter my bias :). But I think it's not *that* hard to reason about, maybe a tooltip telling you exactly that would be enough? User491 even unprompted tried that: _"Trying to set the maximum number of cards to 0 also deosnt work as it just treats this as the default option and ignores it"_.

Just keeping in mind the Hick's Law:

> The time it takes to make a decision increases with the number and complexity of choices.

greg19:
> > `maxAmount` = 0 is not the same as Missing cards. A card is missing if you don't have it in your dex, but you can still have it in your dex but with a 0 amount.
> 
> (Totally didn't know we could have a owned card with zero amount)
> 
> And after testing, we currently have bugs in production regarding 0 amount-owned-cards:
> (changing the amount of Mega Pinsir crash the app)

What is the error in the console? I can't reproduce it

afu-dev:
I've also started the label above the toggle. I'm not a good integrator, so I've only gotten this far for now 😅 

image

greg19 (approved):

PR #870, opened 2026-03-06
Add profile language setting

### LINKED ISSUE #657: I'd like to trade in a specific language only
Make a new trade setting: tradingLanguage with the default being "All languages". Someone can change their setting to a specific language. When you do, the trading matches page will only suggest people that also set their trading language to the same language.

This way you can easily find others that want to trade cards only in a specific language.

greg19:
Closes #657. Adds an optional language setting to user profile. I will also add a dropdown to the trade `/trade/matches` page so you can filter by it.

image

image

image

image


greg19:
@marcelpanse can you try deploying the deno function and test if it works correctly? I've set my trading language to English, you can check if I'm the only one that shows up.

PR #886, opened 2026-04-04
feat: migrate auth service from Supabase to PocketBase

franciscasao:
Replace OTP email login with PocketBase email/password auth. Create PocketBase client lib. Update User type, Login component, authService, and useAuth hooks.


marcelpanse:
Some information and background about what this does and why you would want this would be appreciated

franciscasao:
Sorry I PR'd on the wrong repository locally. I thought I already replaced the origins after forking.

PR #891, opened 2026-04-12
Add missing Promo B cards

WickiRiama:
Add missing Promo B cards


greg19:
Thanks! I forgot to say that you also need to run `pnpm run hashes` to update the card scanner database. It won't be needed in the future though (I've added it to the `scanner` command in #890)

WickiRiama:
Thanks,

Is the script supposed to add the cards images for all languages ? Because I have only the English images since the Fantastical Parade Expansion.

And is it supposed to add locales keys for the new cards ? Because there are missing keys for mega pokemon and trainers cards, also since the Parade Expansion.

greg19:
The images from the other languages were contributed manually by someone else. As of the text translations, you can check sync-card-translations.js, but I'm not sure if it's up to date

PR #894, opened 2026-04-13
Overview table

greg19:
As there are more and more expansions, checking the percentage of a new card requires more and more scrolling and looking for the numbers. Also see [here](https://community.tcgpocketcollectiontracker.com/t/sort-all-sets-packs-by-percent-chance-of-a-new-card/7549).

This PR changes the carousel interface for a table, which can (almost) fit on a monitor screen. Also the fact that the percentages are in one column makes it easier to scan through with your eyes.

In the future, we can introduce an option to sort the packs in the table by the percentage.

image
image


marcelpanse:
Looks nice, I like the overview. Small bits of feedback about alignment and remove the word "cards" since it is also in the column title:
Localhost 2026-04-13 at 3 34 39 PM

PR #899, opened 2026-04-20
Amount wanted

### LINKED ISSUE #611: Override "wanted amount" for specific cards
There were multiple voices, that people want more fine grained control, including use cases as:
1. collecting 5 mewtwo/charizard/... for the missions
2. some people don't want to use the value of 2 for all cards, only those they actually use in the decks

I propose to include a new column in the `card_amounts` table:
```
amount_wanted: number | null
```
This will give an opportunity to override the amount you want to have for specific cards. It would influence both min and max to keep things simple. Value of `null` means fall back to the rarity setting.

In terms of UI I think the best place for this would be the card details side panel. Just remember to make it easy to go back to `null`. There should also be a list of all overriden cards under `Trade/Settings` and an option do quickly discard (set to `null`) them.

I think this is quite neat:
- no additional overhead out of the box
- gives really fine grained control for those who need it (e.g. use case 1)
- packs nicely in the UI

greg19:
Fixes #611.

### DIFF
--- a/frontend/src/components/Card.tsx
+++ b/frontend/src/components/Card.tsx
@@ -36,7 +36,7 @@ export function Card({ card, onImageClick, className, editable = true }: CardPro
   const handleMinusButtonClick = () => {
     if (card.collected && amountOwned === 0) {
       startTransition(async () => {
-        await deleteCardMutation.mutateAsync({ cardId: card.card_id })
+        await deleteCardMutation.mutateAsync({ internal_id: card.internal_id, cardId: card.card_id })
       })
     } else {
       updateCardCount(amountOwned - 1)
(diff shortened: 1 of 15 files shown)


greg19:
@marcelpanse can you add a nullable `amount_wanted` column to the `card_amounts` table?

marcelpanse:
> @marcelpanse can you add a nullable `amount_wanted` column to the `card_amounts` table?

done

greg19:
# Input form

The input spinner is on the `CardDetail` panel

### Value set

image

### Fallback indicator

image

# Trade settings

Additionally, there is a list of all overridden cards in the `Trade/Settings` page 

### List of all custom amounts

image

### Empty list

image

greg19:
@marcelpanse I tested the frontend logic and it looks like it's working. Now, the backend functions need to be deployed and tested

marcelpanse:
> @marcelpanse I tested the frontend logic and it looks like it's working. Now, the backend functions need to be deployed and tested

I deployed the function

greg19:
Okay, it should be working now. I've managed to make `collectionService` work, but it required a lot of effort, maybe it deserves a refacor.

greg19:
BTW, remember to add the column to the public collection view

PR #931, opened 2026-05-10
Trade total changes

### LINKED ISSUE #789: View pending trades when on a person profile (2 features)
## Is your feature request related to a problem? Please describe.

Currently, I'm using (and maintening) the french community post to exchange french card only (link at the end).

I have a few trades pending with a few people and I want to keep at least 2 copy of each cards, here's two problems I'm encountering:
1. I don't want to send a duplicate trade to someone.
2. I don't want to send multiple trade of the same card I own to multiple people (reducing my card count, after an exchange, to something under 2 copy) and also I don't want to waste a trade to get too many copy of a card.

Currently, I have to switch between my trade summary page and the trade page of multiple people to make sure that I don't over-trade or send a duplicate trade.

## Describe the solution you'd like

For the first problem, I imagine a "currently pending trades" on the people trade page. For example, here's the trade that I have with someone:

Image

I would like, when I'm on Bolf trading page, to have a small table to summarize the currently pending trades, like this:

Image

Note: this could/should be a lot smaller/different to avoid confusion.

For the second problem, I imagine a small icon on card that I'm currently trading with a tooltip that shows the pending trades for a card.

Example, I'm currently trading A4b-171 Giratina (2 -> 1) with Nacinra and I don't want the count to reach 0. On NeoGo page, I'm able to trade Giratine again for other cards, but I don't want to.

Image

Image

## Describe alternatives you've considered
I think there are many other alternative, but here are already some ideas to enhance the user experience.

## Additional context
Link to the french community post: https://community.tcgpocketcollectiontracker.com/t/echangeons-nos-liens-pour-le-nouveau-systeme-dechange/6699

greg19:
Closes #789.

On the trading page, the card amount increments (the arrows with from the old to new amount) now show the total changes across all trades.

Please do:
```sql
CREATE OR REPLACE FUNCTION trades_total(id text) RETURNS TABLE(card_id text, count bigint) AS $$
    SELECT
        card_id,
        SUM(count) as count
    FROM (
        SELECT offer_card_id as card_id, -1 as count FROM trades WHERE offering_friend_id = id AND status IN ('offered', 'accepted') UNION ALL
        SELECT receiver_card_id as card_id, 1 as count FROM trades WHERE offering_friend_id = id AND status IN ('offered', 'accepted') UNION ALL
        SELECT offer_card_id as card_id, 1 as count FROM trades WHERE receiving_friend_id = id AND status = 'accepted' UNION ALL
        SELECT receiver_card_id as card_id, -1 as count FROM trades WHERE receiving_friend_id = id AND status = 'accepted'
    )
    GROUP BY card_id
$$ LANGUAGE SQL;

CREATE INDEX ON trades (offering_friend_id, status) INCLUDE (offer_card_id, receiver_card_id);
CREATE INDEX ON trades (receiving_friend_id, status) INCLUDE (offer_card_id, receiver_card_id);
```

image

image


marcelpanse:
1. I don't really like the x2 > x1 you are using on this list because we already use that in various places where it means "when you do this action your card count goes from 2 to 1". In this case it means something else "you have a pending trade", which is very confusing. Instead I suggest using something else like a small orange dot or icon indicating "warning you have a pending trade on this card". When you hover over the icon, it can show a popover explaining what is happening. That should be more clear. 

2. I'm also not a fan of the postgres procedures on the db, they are not in code and are invisible unless you query the database specifically to see what is on there, which makes it really hard to understand the code when you are reading it - it will refer to a 'magic' function on the db. Instead, why not just use the trade-offers list we already fetch in the client and cross-reference on that list on the client? That's more consistent with how we handle it in the rest of the codebase.

greg19:
> I don't really like the x2 > x1 you are using on this list because we already use that in various places where it means "when you do this action your card count goes from 2 to 1". In this case it means something else "you have a pending trade", which is very confusing. Instead I suggest using something else like a small orange dot or icon indicating "warning you have a pending trade on this card". When you hover over the icon, it can show a popover explaining what is happening. That should be more clear.

Other than trading, we use it in scanner, for which it is also a *combined* amount. I think also having *combined* change in trading does make sense. The meaning is similar "when you do [all] your trades, your card count goes from 2 to 1".

I guess you are mainly concerned that it suggests that, for example, my trade with Joy suggests that *completing this trade* will decrease the count by 2? This is different than how it currently displays, but I think it won't be hard for users to switch the thinking to the *combined* approach (but yes, they will have to adapt).

I think the *combined* approach is more *useful*. Like, everyone knows that completing a trade decreases your count by one, and it is so simple math that the usefulness of subtracting one by the UI is rather low (actually, the only reason I initially included it is to distinguish the card you are giving from the card you are receiving; now we have the chevrons which do it better). It is more useful than a dot, as you can directly see if you want to trade or not: if you have 3 copies, and have 2 pending trades, you don't; but if you have 3 copies and 1 pending trade, you can.

I don't think we need to babysit the users so much, they might be *slightly* confused, but IMO they will catch up rather quickly. And people actually derive value from a pragmatic UI that doesn't require 6 clicks to do a single thing (or the other way around: they become frustrated when claiming daily mission requires 6 click with 700ms delay after every tap – looking at you DeNA).

> 2. I'm also not a fan of the postgres procedures on the db, they are not in code and are invisible unless you query the database specifically to see what is on there, which makes it really hard to understand the code when you are reading it - it will refer to a 'magic' function on the db.

Are you open for using supabase migrations for future changes in the codebase? AFAIK that's what they recommend to solve the exact problem you are raising. I can create the migration for this PR if you want.

Alternatively, we can just `pg_dump` it. This problem is also present for example for the database schema. Having the schema (and the custom functions) `pg_dump`ed will really improve the possibility of first time contributions IMO (I remember that I did not know what is supabase and was confused how can I change something in the database).

> Instead, why not just use the trade-offers list we already fetch in the client and cross-reference on that list on the client? That's more consistent with how we handle it in the rest of the codebase.

Databases are fast, transferring larger objects over the network and doing calculations in single threaded JS is slower. I understand that here, it won't be felt by the user, as it's like 40 objects. Also we do already fetch the list on the offers page, but that is not the case for the TradeWith page.

I think my general point is that "when something *can* be done on the client side, it means that it *should* be done on the client side" is not the good rule of thumb. I much prefer the rule "fetch exactly the data you need, and allow the database to do the heavy lifting". Also, I just don't see the appeal of replacing 4 line SQL with 10 lines of JS. Maybe it's a skill issue on my end, but I don't think I would be able to write the JS logic anywhere close to the readability of the SQL.

Somewhat of a tangent, but the fact that some parts of the site take multiple seconds to load sometimes frustrate me. I just measured that simply opening the trade offers page takes 1s to load downloading 4MB of data over 100 network requests. Really, why do I have to download 500kB of xlsx.js *before* I can start fetching the trades?!

greg19:
After some more thinking, I think I'm okay with using the existing active trades object to do the calculations in the frontend.

But, for my workflow, I'm still not interested in any solution that doesn't show the amount after all trades without requiring additional clicks.

marcelpanse:
> After some more thinking, I think I'm okay with using the existing active trades object to do the calculations in the frontend.

Great, that makes it at least consistent with the codebase.

> But, for my workflow, I'm still not interested in any solution that doesn't show the amount after all trades without requiring additional clicks.

What if we just show something like: A1 Bulbasaur   (1) x3
where the (1) is an orange circle (like Tailwind CSS Badges - Official Tailwind UI Components 2026-05-14 at 8 16 37 PM) with a one it showing you have (1) already in either pending or accepted state. If you hover over it we can show a tooltip explaining it in words. 

https://tailwindcss.com/plus/ui-blocks/application-ui/elements/badges

That should make it obvious how many you have 'in trading' and also clear how it works?

greg19:
image

PR #933, opened 2026-05-11
Deck update time

greg19:
Currently the decks with the most likes are really old. This PR adds last update time to the decks list so you can see which have been recently updated and which did not.

Please add the `updated_at` column to the `public_decks` view.

- image
- 
image


marcelpanse:
Any deck that is not your own shows an error in the date:
Localhost 2026-05-11 at 9 15 12 PM

Localhost 2026-05-11 at 9 15 40 PM

greg19:
> Please add the updated_at column to the public_decks view.

I thought it was the issue with the view, isn't it?

marcelpanse:
It seems it was not. Just added it, seems good now 👍

PR #952, opened 2026-05-23
Cleanup toggle & tabs filters

greg19:
- Removed dead code
- Adjusted colors
- Fixed a bug where rarities could not be selected with keyboard

### Before

image

### After

image


greg19:
Also, we should use `<Toggle type="single">` instead of the tabs filter. There is no `<TabsContent>` so tabs don't really make sense. But that's for another PR

PR #962, opened 2026-06-02
Tweak trade matches algorithm

greg19:
There is a big problem that people do not accept pending trades (I didn't dig too deep for the reason, but my guess is because they just don't log in to the site). I added a few additional conditions trying to improve this.

1. `collection_last_updated >= now() - interval '4 days'`. Additional safe guard to filter inactive users. Mostly applicable to single card query, which can have significantly lower base number of users. But can also affect less popular languages.
2. `accepted trades < 8`. Give the other person some breathing room to complete their accepted trades. If they have a lot of them, it's rather unlikely that you will make the trade anytime soon. Even if they accept, you will just see a permanent "Trade in progress" in the app.
3. `received trades < 3`. First reason, do not overflow the person with too many requests when they next log in, and possibly route the requests to the next person on the list, resulting in more balanced distribution of pending trades. Second, they might be not accepting their requests because they might be in fact inactive (which should be caught by the condition 1. soon, but this condition might be quicker).

For a reference, there are about 230 users which updated their collection within last 24 hours. The `LIMIT 50` captures only those who did in the last 3 hours. Out of the 230 users, 10 had `accepted trades >= 8`, with the highest being 35. Out of the 230 users, 25 had `received trades >= 3`, with the highest being 48 (!).

### DIFF
--- a/frontend/src/pages/trade/TradeMatchesResults.tsx
+++ b/frontend/src/pages/trade/TradeMatchesResults.tsx
@@ -64,7 +64,7 @@ export default function TradeMatchesResults() {
   }
 
   if (data.length === 0) {
-    return <p className="text-xl text-center py-8">{t('noTradePartners')}</p>
+    return <p className="text-lg text-center px-2 py-8">{t('noTradePartners')}</p>
   }
 
   const friendIdSet = new Set(friends.filter((f) => f.state === 'accepted').map((f) => f.friend_id))
--- a/scripts/sql/trade-partners.sql
+++ b/scripts/sql/trade-partners.sql
@@ -9,6 +9,9 @@ WITH recent_accounts AS (
       AND username IS NOT NULL
       AND collection_last_updated IS NOT NULL
       AND ($2::text IS NULL OR language = $2::text)
+      AND collection_last_updated >= now() - interval '4 days'
+      AND (SELECT COUNT(*) FROM trades WHERE status = 'accepted' AND (offering_friend_id = friend_id OR receiving_friend_id = friend_id)) < 8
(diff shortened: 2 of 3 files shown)


marcelpanse:
I'm conflicted about this. It makes sense to get the best possible offers. However, I don't get that many trade options in the first place, I only have 5-10 people at the moment to trade with, so I'm happy to find anyone with matches, I don't want to restrict that even further and possibly miss out on valid trades.

I think the first option makes sense, and won't have too much impact since they are already sorted on the most recent active 50 people anyway.
The other two would restrict my possible trades list further. Could we solve that a different way?

What I'm also thinking, but that's unrelated to this ticket, is that it would greatly increase my trading speed if I just would know when someone in-game accepted my trade, or offered a card. Now I only log in twice a day, but if I would have a notification that something happened I could act immediately. Since the game does not send you notifications for this, we should build something instead. If everyone would be trading faster, then this would be less of an issue. For example, we could send (opt-in) emails about trading status updates. It would also help with the unresponsive people not accepting trades.

greg19:
> The other two would restrict my possible trades list further.

It doesn't restrict, it just balances it in a different way. It still considers 50 users, but replaces some with the next person on the list (who wouldn't be considered for trade without these changes). And I don't see why the next person (who updated their profile 4 hours ago, previously not considered by the LIMIT 50) would be a worse trading partner (in expectation) than a person who updated 2 hours ago (and additionally has 35 trades they did not respond to).

greg19:
The whole deal with the two later filters is to *increase* your chance of finding a person, by including users further down the list at the cost of the ones that are statistically less likely to accept your trades.

marcelpanse:
Then I read it wrong, I thought it would filter the list.

PR #976, opened 2026-06-24
Empty give card in trade

### LINKED ISSUE #648: Allow sending trades with empty card to give
I'd be cool to have an option to propose a trade and only saying what card you want to receive, without saying what you want to give. I guess many people don't care about the card they give, and it would allow the receiver to choose his favorite one. This could increase the fraction of proposed trades to actually finished ones, because both people would actually be truly interested in the trade. Currently I can think the receivers often have "yea, I guess I can do it" instead of "This is what I've been looking for!", resulting in many stale trades (at least I've got many trades "accepted" but the trade actually not happening in game).

This tackles similar issue to #625, but I think is much neater (in my opinion), will improve UX, and is easier to implement (no need to introduce yet another state).

greg19:
Closes #648.

- Adds a QoL feature that TradeWith shows friend header before the full friend collection is loaded
- A new `SelectCard` dialog that can be later (in a separate PR) used in other places (like the TradeMatches.tsx, from which most of it was copied from)
- I changed some services to export `queryOptions` instead of `useQuery` result. The reason was that I needed to fetch the friend collection after the button click, and it would cumbersome to do it with a hook inside the component, as the `enabled` management would be tedious. We don't want to prefetch it when you load the page, as the query is very heavy, taking 1-3 seconds. So I used `queryClient.ensureQueryData` to get the data on click. For reusability of the query params I changed some `useFoo(arg)` into `useQuery(fooQuery(arg))` so that it can also be used with `queryClient.ensureQueryData(fooQuery(arg))`.

### Offerer point of view

You can now offer a trade without specifying your card:
image

If you offered the trade without a card, you can fill it even before the other person responds:
image
image
If you do, the trade will start behaving as a trade with both cards filled that you would offer.

If you leave the card empty, the other person can choose it and you will see it as an offer from them:
image
From now onward, this trade offer behaves the same as if the other person would send you an offer with both cards selected.

### Receiver point of view

When you receive such offer you can choose the card you want:
image
image

After you select the card, the original sender needs to accept it:
image
From now onward, this trade offer behaves exactly the same as an offer you would make with both cards selected.


greg19:
@marcelpanse Can you:

```sql
ALTER TABLE public.trades
  ALTER COLUMN offer_card_id DROP NOT NULL,
  ALTER COLUMN receiver_card_id DROP NOT NULL;
```

marcelpanse:
> @marcelpanse Can you:
> 
> ```sql
> ALTER TABLE public.trades
>   ALTER COLUMN offer_card_id DROP NOT NULL,
>   ALTER COLUMN receiver_card_id DROP NOT NULL;
> ```

Done

PR #988, opened 2026-07-29
Common collection state for expansion

greg19:
We are finally getting it!

> This expansion comes with a few important updates to core game features:
> - When you obtain a card included in multiple booster packs, it will be registered in the card dex for all corresponding booster packs. This change will also apply to cards you already own.

It will come with the next pack.

@marcelpanse Please do:
```sql
ALTER TABLE card_amounts
ADD COLUMN collected boolean NOT NULL DEFAULT false;

UPDATE cards_amounts AS ca
SET collection = EXISTS (
    SELECT 1
    FROM collection AS c
    WHERE c.email = ca.email
      AND c.internal_id = ca.internal_id
);
```

So I can test the PR.


marcelpanse:
I ran the SQL but only for your email specifically so you can test it.

frontend/src/services/collection/collectionService.ts
```diff
@@ -186,34 +169,29 @@ export async function updateAmountWanted(
...
-    if (internal_id === undefined) {
-      console.error('Invalid card_id:', cardId)
-      continue
-    }
+  for (const internal_id of internal_ids) {
     const row = collection.get(internal_id)
-    if (row?.collection.includes(cardId)) {
-      row.collection = row.collection.filter((id) => id !== cardId)
+    if (row?.collected) {
+      row.collected = false
```
marcelpanse: @greg19  this should be `row.collected = collected`?

greg19:
I tested the collection page, card details and scan and they work

marcelpanse:
@greg19 
1. I ran the SQL on my account, but the cards still showed as 'uncollected' because of the localstorage cache. We need to invalidate and make a v4 cache. Can you adjust the cache name to force this?
2. the sql took ~3sec to run on just my account. Not sure if this will be feasible to run on ~25k accounts. If not I'll need to make a batch script to do it in parts instead. I'll do this right before deployment.
3. setCollected uses update(), not upsert, rows in card_amounts are created lazily, so ticking the new checkbox on a card the user never touched matches 0 rows,silently "succeeds", and then writes the optimistic local cache + bumps its timestamp. The state looks collected but is nowhere in the DB. If you clear the localstorage cache and refresh, the collected card is uncollected again. Can you fix?

greg19:
> 1. I ran the SQL on my account, but the cards still showed as 'uncollected' because of the localstorage cache. We need to invalidate and make a v4 cache. Can you adjust the cache name to force this?

The PR already bumps it from v2 to v3. You might have opened the PR before you ran the SQL for yourself, in this case just clear the cache in your browser.

> 3. setCollected uses update(), not upsert, rows in card_amounts are created lazily, so ticking the new checkbox on a card the user never touched matches 0 rows,silently "succeeds", and then writes the optimistic local cache + bumps its timestamp. The state looks collected but is nowhere in the DB. If you clear the localstorage cache and refresh, the collected card is uncollected again. Can you fix?

Fixed.

Issue #969, opened 2026-06-18
Deckbuilding mode for trading


marcelpanse:
So basically “deck-building mode for trade” so it collapses all alternative cards into one when considering the trade amounts.

It would be cool, if possible, to add an option similar to the one in the overview page where you can only see copies for deck-building regarding of the rarity for the trading settings.

For example i currently own at least 1 copy of every card up untill 1 star rarity, and right now i’m only intereseted in finding trade for missing duplicates from 1 to 4 diamond rarity, of course many times based on other promo card that i have or having luck on finding the same card but with higher rarity i receive many trade that i mannually check to see if i have just 1 copy of that card in deck building mode.

It’s not that i don’t want to do that but it could help other people to select more fine grained trade to propose/select.

nikshitak:
I agree, I think this issue would be really useful and I'm willing to work on it. 

Just to get logistics out of the way:

I'm using the following example:
If I have one 1 star shiny Gholdengo with the move Scintillating Surfing from Paldean Wonders and 1 3-diamond Gholdengo with the same move, then this should count as 2 cards for deck building purposes, but ALSO, in the Overview, when I select "Deck-building mode" and have "Number of cards wanted" as 2, then it should count both of these cards as the same card, and take this into consideration when doing the math for what pack to open next. 

It also just makes things easier to wishlist, especially if, because of this feature, you can also see what cards you *truly* don't have ≥ 2 of, and can wishlist those on the app/trade for on the website.

marcelpanse:
```but ALSO, in the Overview, when I select "Deck-building mode" and have "Number of cards wanted" as 2, then it should count both of these cards as the same card, and take this into consideration when doing the math for what pack to open next.```

Yes that's already how it works currently so you can take that as the base and reuse/reapply the logic on the trading part

The only difficult part here is that it should affect the trade matches, which means it has to be done on the backend function in an already pretty complex SQL query.

nikshitak:
Oh I see - I couldn't see what cards specifically it thought I didn't have, so I was confused whether that was the logic or not. Would that be a separate issue? 

Also, when you say that it should affect trade matches, does that mean it should consider all cards until the user only has one card of that rarity left (1 card if the card pops up in multiple sets/rarities, 2 cards if it doesn't)?

marcelpanse:
[...]

PR #995, opened 2026-08-05
Carddex trading

greg19:
I wanted to make this stacked on top of #988, but I didn't realize you need GitHub CLI for the new stacked PRs. I'll just fix git manually after #988 is merged.

image

This PR allows you to tick a "Collecting CardDex" option for trading. The option *additionally* considers the card as wanted if you don't have it registered in your CardDex.

It was requested by a user on the forum, who only cares about the CardDex and allows trading his cards away to 0 with options `to_collect: 1, to_keep: 0`. He was frustrated that the site considers the cards as "wanted", even though he does not want them (cards that are registered, but with 0 copies). This option addresses this issue when parried with `to_collect: 0` (or individual card `amount_wanted: 0`).

Right now it is a rather niche option, but will be much more useful when paired with #969. Then, you will be able to simultaneously collect card dex and deck build, where trading down to 0 seems plausible (when you have 2 full art cards).


greg19:
@marcelpanse Can you:

```sql
ALTER TABLE trade_rarity_settings
ADD COLUMN collecting_carddex boolean NOT NULL DEFAULT false;
```

marcelpanse:
> @marcelpanse Can you:
> 
> ```sql
> ALTER TABLE trade_rarity_settings
> ADD COLUMN collecting_carddex boolean NOT NULL DEFAULT false;
> ```

done

greg19:
I ended up placing the option in the "to collect" column, as it defines when a card is considered wanted, and it doesn't make any sense with `to_collect` value other than 0 (a positive value already implies that uncollected cards are wanted).

image

marcelpanse:
@greg19 can you check the merge conflict so I can test it and merge it?

PR #649, opened 2025-09-21
feat: add A3b cards for all languages and translations for tools and trainers

mpierrez:

PR #1002, opened 2026-08-29
fix(ui): stop image flicker loop and use locale image in card modal

diegomarnie:
Two related bugs in card image display:

1. FancyCard onError fallback wrote to DOM directly via (e.target as HTMLImageElement).src = card.image, but the JSX kept src={imagePath}. Any re-render (hover, mouse move throttled to 20fps) reconciled the DOM back to the localized path, re-triggered onError, and looped. Very visible on P-B-80..86 (no localized version, so fallback runs every render).

   Fixed by tracking imgSrc in state so the fallback survives re-renders.

2. The big-image modal opened from the card detail sheet hardcoded card.image, which is always the /images/en-US/ path — so opening the modal always reverted to the English scan even if the tile above showed the localized one.

   Extracted CardImageWithFallback in CardDetail.tsx using the same state-based locale-first / English-fallback pattern.


greg19:
I'd implement the localization properly from scratch instead of patching the current approach.

As you noticed, it's too tempting to just use `card.image` without the localization function call. So I'd just remove this property. It shouldn't be used for localization anyway I believe?

I believe the simplest approach of solving the issue we are constantly facing with patching the `src` when a fetch fails is to make it never fail.
- The best option is to research from where the localized cards were taken, and incorporate downloading them in the scraper. I'd really put an effort in that, as this would solve all the problems.
- If you are not able to do that, then I'd symlink the localized versions to the English ones.

This way we never need to worry about any fallbacks, and there is no need for any code to handle that (which, as you see, can lead to bugs).

BTW, while we are on the image topic, I'd change the file names to use internal ids instead of card ids, to not waste any resources (disk space, network bandwidth, ...)

diegomarnie:
Thanks for the direction — you're right that this is a patch, not a real fix. Before restructuring I dug into where the localized images could come from and wanted to share what I found:

**Public sources I checked (all English-only):**
- `pocket.limitlesstcg.com` CDN: `B4a_001_EN.png` returns 200, every `_ES`/`_FR`/`_DE`/`_IT`/`_PT-BR` variant 403. Their own `?lang=es` pages still render the `_EN` file.
- `pocket.pokemongohub.net/tcg-pocket/cards/{set}/webp/{n}_{lang}.webp`: only `_en` returns 200, and not for every set (B3, P-B 404 even in EN).
- Community repos I found via GitHub search — `chase-mew/pokemon-tcg-pocket-cards`, `alevalls99-maker/ptcgp-assets`, `hugoburguete/pokemon-tcg-pocket-card-database`, `flibustier/pokemon-tcg-pocket-database`: only English images.

**Interesting lead:** `flibustier/pokemon-tcg-pocket-database` has metadata in multiple languages (`dist/cards.fr.json`) and their `scripts/download-images.ts` supports all 9 game languages, but the actual endpoint is behind a private env var (`ASSETS_CARD_IMAGES_ENDPOINT`) and the language block is commented out. Image filenames use the raw game asset names (`cPK_10_000010_00_FUSHIGIDANE_C.webp`), which strongly suggests APK datamining.

**Existing localized images in this repo** are 367×512 WebP RGBA — the game's native asset resolution — matching that datamining theory. They stopped being added around A4a (looking at the commit history: `add [set] cards for all languages` PRs #334, #366, #440, #649, #650, #655, #805, #806).

So the "research the source" path realistically lands on one of:
1. Reach out to flibustier and ask about the endpoint (uncertain response time).
2. Extract from the APK ourselves via `Layen-lang/PTCGP-Aladin-decryptor` (technical but reproducible per patch).
3. Watch community dumps (Discord/Reddit — variable quality).

I don't want to force a half-baked answer here. A few options for this PR:

- **A.** Close this PR. Open a fresh one once we've pinned down one of the three sources above, doing the full change you described (remove `card.image`, do the download in the scraper for all locales, rename files to `internal_id.webp`).
- **B.** Keep this PR narrowed to just the flicker fix in `FancyCard.tsx` — that loop was hitting ~20 failed loads per second on any card without a localized image, and would still be worth landing even if we redesign later. Drop the modal change.
- **C.** Skip A/B for now and I do the APK-extraction route to actually solve it.

Happy to go whichever way you prefer. Let me know.

greg19:
Okay, I see that getting the assets will be hard. What about just copying the English versions to the localized directories in place of the missing ones (I previously suggested symlinks but now I read that GitHub pages might have some issues with them)? Then we don't need any fallback, so no flickering.

diegomarnie:
Before I go do the copy, one alternative worth considering — happy to drop it and do the copy if you prefer:

**Cost of the copy approach:** I ran the numbers on `main`. Currently missing localized files vs `en-US`:

| locale | missing | size |
|---|---|---|
| es-ES | 2093 | 109 MB |
| fr-FR | 1383 | 73 MB |
| de-DE | 1677 | 89 MB |
| it-IT | 2093 | 109 MB |
| pt-BR | 2093 | 109 MB |
| **total** | **9339 files** | **~490 MB** |

That would nearly double the repo (~582 MB → ~1.07 GB) and every new set adds another ~500 files of duplication.

**Alternative:** precompute a small `image-locales.json` manifest at build time (`{card_id: ['es', 'de', ...]}`), and have `FancyCard` read the correct path from that map. Same guarantee as the copy — the decision is made ahead of time, no runtime `onError`, no flicker — but no bytes duplicated. A new set updates the manifest via the same `pnpm hashes`-style step (the localized folders already exist, just enumerate them).

Roughly ~15 lines of build-time code + a ~30 KB manifest replaces 490 MB of duplicated files.

If you'd rather stick with the copy for simplicity's sake I'll do that instead — just wanted to flag the tradeoff.

marcelpanse:
I prefer not copying the card literally because it also makes it hard to see which cards are missing in which languages. A generated json at build time makes more sense.

greg19:
Yeah, makes sense

diegomarnie:
Force-pushed the manifest approach.

**What's in it**

- `scripts/generate-image-locales.ts` — walks `frontend/public/images/*/` and writes `frontend/assets/image-locales.json`: `{card_id: bitmask}` over `[en-US, es-ES, fr-FR, de-DE, it-IT, pt-BR]` (positional — new locales appended to keep bits stable). Manifest is ~60 KB for 3873 card ids.
- Added it to the `pnpm scraper` chain so the manifest is regenerated any time images move.
- `frontend/src/lib/imageLocales.ts` exposes `getLocalizedImagePath(card, lang)` (and a `hasLocalizedImage` helper). Handles the short-code → BCP-47 mapping (i18next gives `es`, folder is `es-ES`, etc.).
- `FancyCard`, `CardDetail`'s big-image dialog and `DeckView` now use the helper for `src` — no `onError`, no flicker, and the modal now opens the localized image when one exists (was hardcoded to `card.image`).
- `CardDetectionService.getRightPathOfImage` (the async `` probe) is replaced by the sync lookup.

**Net effect vs the copy approach**

- +60 KB manifest instead of +490 MB of duplicated files.
- New sets pick up automatically once the scraper runs (which already regenerates the manifest).
- Missing localizations remain visible in `image-locales.json` — you can see at a glance which cards lack which locales, per @marcelpanse's point.

`pnpm build` passes.

I did not remove the `card.image` property from `cards.json` — happy to do that as a follow-up (would need the scraper change plus removing the remaining reads outside the ones this PR touched), but wanted to keep this PR focused.

PR #1005, opened 2026-08-29
feat: add missions for B3a, B3b, B4 and B4a

diegomarnie:
## Summary

The mission scraper's expansions list stopped at `B3`, so mission data for the four newest sets was never generated. `B3a-missions.json` existed but was empty.

Added the four sets to `scripts/scrape-missions.js` and generated their mission files from pokemon-zone.com:

- **B3a**: 14 missions
- **B3b**: 11 missions
- **B4**: 15 missions
- **B4a**: 19 missions

Also registered the three new JSONs in `frontend/src/lib/MissionsDB.ts` — the scraped files alone don't surface in the UI without an entry in `missionsMap`.

## Test plan

- [x] `pnpm scraper` (missions run) produces the four JSONs cleanly
- [x] Frontend shows the four sets under Missions with expected counts
- [x] `pnpm build` passes


scripts/scrape-missions.js
```diff
@@ -7,7 +7,7 @@ const TARGET_DIR = './frontend/assets/themed-collections/'
 const CARDS_PATH = './frontend/assets/cards.json'
 const UA = 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0 Safari/537.36'
 
-const expansions = ['A1', 'A1a', 'A2', 'A2a', 'A2b', 'A3', 'A3a', 'A3b', 'A4', 'A4a', 'A4b', 'B1', 'B1a', 'B2', 'B2a', 'B2b', 'B3']
+const expansions = ['A1', 'A1a', 'A2', 'A2a', 'A2b', 'A3', 'A3a', 'A3b', 'A4', 'A4a', 'A4b', 'B1', 'B1a', 'B2', 'B2a', 'B2b', 'B3', 'B3a', 'B3b', 'B4', 'B4a']
```
greg19: Can we instead `import { expansions } from '../frontend/src/lib/CardsDB'`? So the missions scrape automatically after adding a new expansion.

frontend/src/lib/MissionsDB.ts
```diff
@@ -38,6 +41,9 @@ export const missionsMap: Partial<Record<ExpansionId, Mission[]>> = {
   B2b: B2bMissions.default as Mission[],
   B3: B3Missions.default as Mission[],
   B3a: B3aMissions.default as Mission[],
+  B3b: B3bMissions.default as Mission[],
+  B4: B4Missions.default as Mission[],
+  B4a: B4aMissions.default as Mission[],
 }
```
greg19: Can you figure out how to change the code in a way that we don't need to add these lines every time?

frontend/src/lib/MissionsDB.ts
```diff
@@ -18,6 +18,9 @@ const B2aMissions = await import('../../assets/themed-collections/B2a-missions.j
 const B2bMissions = await import('../../assets/themed-collections/B2b-missions.json')
 const B3Missions = await import('../../assets/themed-collections/B3-missions.json')
 const B3aMissions = await import('../../assets/themed-collections/B3a-missions.json')
+const B3bMissions = await import('../../assets/themed-collections/B3b-missions.json')
+const B4Missions = await import('../../assets/themed-collections/B4-missions.json')
+const B4aMissions = await import('../../assets/themed-collections/B4a-missions.json')
```
greg19: Can you figure out how to change the code in a way that we don't need to add these lines every time?

diegomarnie:
Refactor pushed. Both hardcoded lists are gone.

- `scripts/scrape-missions.js` → `.ts` and imports `expansions` from `frontend/src/lib/CardsDB`, deriving the id list with `.map((e) => e.id)`. `pnpm scraper` runs it through `tsx` now.
- `frontend/src/lib/MissionsDB.ts` uses `import.meta.glob<{ default: Mission[] }>('../../assets/themed-collections/*-missions.json', { eager: true })`, matches each path to a known `ExpansionId`, and builds `missionsMap` from that.

Adding a future set is now zero-code: `pnpm scraper` produces the JSON and the frontend picks it up.

Build & biome pass.

PR #1006, opened 2026-08-29
fix: correct B4a trainer rarities

diegomarnie:
## Summary

Four trainers in the recently added Team Rocket's Ambition (B4a) set were stored with incorrect rarities compared to Limitless.

| Card | Was | Should be |
|---|---|---|
| B4a-67 Team Rocket's Thieving Machine | ◊ | ◊◊ |
| B4a-68 Team Rocket's Goo-zooka | ◊ | ◊◊ |
| B4a-69 Team Rocket's Researcher | ◊ | ◊◊ |
| B4a-72 Arcade | ◊◊◊ | ◊◊ |

Each mismatch was cross-checked against its individual card page on Limitless (`/cards/B4a/67`, `/68`, `/69`, `/72`).

## Test plan

- [x] `pnpm run hashes --verify` passes locally
- [x] `pnpm run biome:check` passes (pre-commit)
- [x] `pnpm build` passes (pre-push)


greg19 (changes requested):
This is incorrect, the internal ids are not updated, please re-run the scrapper script

marcelpanse:
What greg said. Any manual changes will be overridden by the scraper on next run.

diegomarnie:
Force-pushed the scraper-based version. Ran `SCRAPER_ONLY=B4a pnpm scraper` (added a temporary env-var filter to the loop, then reverted the script change) plus `pnpm hashes`.

The four rarities now come from the scraper directly, and the four `internal_id`s shift accordingly:

| card | rarity | internal_id |
|---|---|---|
| B4a-67 Team Rocket's Thieving Machine | ◊ → ◊◊ | 1380544 → 1380545 |
| B4a-68 Team Rocket's Goo-zooka | ◊ → ◊◊ | 1380608 → 1380609 |
| B4a-69 Team Rocket's Researcher | ◊ → ◊◊ | 1380672 → 1380673 |
| B4a-72 Arcade | ◊◊◊ → ◊◊ | 1380866 → 1380865 |

`frontend/public/hashes/en-US/hashes.json` reflects the renamed keys (same hash values, images unchanged). `pnpm hashes --verify` passes locally.

The diff is bigger than the 4 rarity/id changes because the scraper also picked up other current-Limitless drift for B4a: `set_details` gained an apostrophe (`teamrocket'sambition`), `alternate_versions` cleanup, minor effect/ability text tweaks. All of that would land on your next scrape anyway.

greg19:
I'm still getting a different output from `pnpm run scraper` than what you are submitting.

diegomarnie:
Could you share a snippet of the diff you're seeing between your local `pnpm run scraper` output and this branch? That'd help me tell whether it's live-Limitless drift between our runs, a formatting mismatch (I wrote a custom JSON serializer to preserve inline arrays), or something else. Even a few lines around one specific card would be enough to pinpoint.

greg19:
Do you have any issues with just running the command? There should be no need for a custom serializer, `pnpm lint:fix` at the end of the pipeline should handle it.

diegomarnie:
Dug into this today and found the actual cause.

**Biome's default `files.maxSize` is 1 MB. Our `cards.json` is 2.83 MB.** Biome silently ignores files over that threshold, so `pnpm lint:fix` at the end of `pnpm scraper` never touches `cards.json` at all.

Reproduced locally on `main`:

```sh
$ pnpm biome check --write frontend/assets/cards.json
Checked 1 file in 1117ms. No fixes applied.     # ← skipped, but says nothing about "ignored due to size"

$ pnpm biome format --write --files-max-size=10000000 frontend/assets/cards.json
Formatted 1 file in 406ms. Fixed 1 file.       # ← now it actually runs

# Before (multi-line, what JSON.stringify produces):
        "cost": [
          "grass",
          "colorless"
        ],

# After (inline, what's in main today):
        "cost": ["grass", "colorless"],
```

That's why my run produced a different output from yours — my raw scraper output kept the multi-line arrays that biome couldn't collapse, so I wrote a Python serializer to match the on-disk `main` format. Yours matches `main` for a different reason (maybe you have a larger biome cap locally, an editor plugin, or ran biome once with the override).

**Suggested fix, orthogonal to this PR:** bump `files.maxSize` in `biome.json` (e.g. to `4000000`). Otherwise every scraper run silently drifts the formatting of `cards.json`, and reviewers see huge stylistic diffs mixed with real data changes. Happy to open a tiny separate PR for that.

**For this PR:** if you'd like, I can re-run the scraper with the size-override applied and let biome format the output natively (no custom serializer). Then the diff will exactly match yours. Want me to do that here or keep it separate?

marcelpanse:
Closing this. I ran this on #1009 with a proper migration script to update the database records.

PR #1008, opened 2026-08-30
feat(collection): add stage/kind/subtype/ability/HP/retreat filters

diegomarnie:
## Summary

Adds six new filters to the collection view, following a community forum request for more precise filtering.

### New filters

| Filter | Values | Applies to |
|---|---|---|
| **Stage** | Basic / Stage 1 / Stage 2 | Pokémon |
| **Pokémon kind** | Regular / ex / Mega ex | Pokémon |
| **Trainer sub-type** | Item / Supporter / Tool / Stadium | Trainers |
| **Ability** | All / Yes / No | Pokémon |
| **HP** | min + max dropdown pair | Pokémon (0 to 350) |
| **Retreat cost** | min + max dropdown pair | Pokémon (0 to 4) |

### Implementation

All filters use existing `Card` fields — no new data or scraping.

- `frontend/src/lib/filters.ts`: added option arrays + types, extended `FiltersAll`, added six predicates in `getFilteredCards`.
- `frontend/src/pages/collection/CollectionCards.tsx`: extended the Zod schema so values persist via URL params like every other filter.
- `frontend/src/components/FiltersPanel.tsx`: rendered with the existing `ToggleFilter` / `TabsFilter` / `DropdownFilter` primitives (no new components).
- `frontend/public/locales/*/filters.json`: added translations for en-US, es-ES, fr-FR, de-DE, it-IT and pt-BR.

### Design notes

- **Mega ex** is detected via `card.ex && card.name.startsWith('Mega ')`. If Pocket ever introduces Mega Pokémon that aren't ex, we'll need a dedicated field, but that's not on the horizon.
- **Retreat cost** filter treats non-Pokémon cards as always matching, so capping "max retreat = 2" still shows trainers alongside the Pokémon that qualify.
- **Ability** looks at both `ability.name` and `ability.effect` — the Trainer text pathway stores its effect on the same `ability.effect` field (name empty), so trainers show as "Yes" if they have an effect. Happy to reshape if you'd rather ability be Pokémon-only.

### Not included (per plan)

- Attack damage min/max (needs parsing of the string `damage` field with variable-damage handling).
- Attack cost (min/max total + energy types).
- Special tags (Ultra Beast / Ancient / Future) — not currently in the `Card` model.

I'd like to land these six first and open follow-ups if you want any of the above.

## Test plan

- [x] `pnpm build` passes (biome + tsc + vite).
- [x] Sanity-checked predicates against real `cards.json`: e.g. `stage=basic|stage1|stage2` sums to 3558 (matches total Pokémon), `kind=megaex` returns exactly the 110 Mega ex cards, `ability=yes` returns 764.
- [x] URL persistence: opening `?stage=basic&ability=yes` restores state.

### DIFF
--- a/frontend/src/lib/filters.ts
+++ b/frontend/src/lib/filters.ts
@@ -1,19 +1,31 @@
 import i18n from '@/i18n'
-import { type Card, type Collection, cardTypes, expansionIds, type Rarity, type RaritySettingsRow } from '@/types'
+import { type Card, type Collection, energies, expansionIds, type Rarity, type RaritySettingsRow } from '@/types'
 import { allCards, getCardByInternalId } from './CardsDB'
 import { levenshtein } from './levenshtein'
 import { getCardNameByLang, getExtraCards, getNeededCards } from './utils'
 
 export const expansionOptions = ['all', ...expansionIds] as const
 export const sortByOptions = ['expansion-newest', 'rarity', 'type', 'recent'] as const
-export const cardTypeOptions = cardTypes
+export const cardTypeOptions = [...energies, 'dragon', 'colorless'] as const
 export const ownershipOptions = ['all', 'missing', 'registered'] as const
 export const tradingOptions = ['all', 'wanted', 'extra'] as const
+export const stageOptions = ['basic', 'stage1', 'stage2'] as const
+export const pokemonKindOptions = ['regular', 'ex', 'megaex'] as const
+export const trainerSubtypeOptions = ['item', 'supporter', 'tool', 'stadium'] as const
+export const abilityOptions = ['all', 'yes', 'no'] as const
+export const hpOptions: readonly number[] = [...new Set([0, ...allCards.map((c) => c.hp).filter((v): v is number => v !== undefined)])].sort((a, b) => a - b)
+export const retreatOptions: readonly number[] = [...new Set([0, ...allCards.map((c) => c.retreat).filter((v): v is number => v !== undefined)])].sort(
+  (a, b) => a - b,
+)
 export type ExpansionOption = (typeof expansionOptions)[number]
 export type SortByOption = (typeof sortByOptions)[number]
 export type CardTypeOption = (typeof cardTypeOptions)[number]
 export type OwnershipOptions = (typeof ownershipOptions)[number]
 export type TradingOption = (typeof tradingOptions)[number]
+export type StageOption = (typeof stageOptions)[number]
+export type PokemonKindOption = (typeof pokemonKindOptions)[number]
+export type TrainerSubtypeOption = (typeof trainerSubtypeOptions)[number]
+export type AbilityOption = (typeof abilityOptions)[number]
 
 export interface FiltersAll {
   search: string
@@ -29,6 +41,14 @@ export interface FiltersAll {
   maxNumber: number | '∞'
   deckbuildingMode: boolean
   allTextSearch: boolean
+  stage: StageOption[]
+  pokemonKind: PokemonKindOption[]
+  trainerSubtype: TrainerSubtypeOption[]
+  ability: AbilityOption
+  minHp: number
+  maxHp: number | '∞'
+  minRetreat: number
+  maxRetreat: number | '∞'
 }
 export type Filters = Partial<FiltersAll>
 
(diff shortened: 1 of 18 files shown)


greg19:
Can you add screenshots?

diegomarnie:
Sure — here's the panel with all six new filters in place (Stage, Pokémon kind, Trainer sub-type, Ability, HP range, Retreat range):

filter panel

And here's the result grid with `Pokémon kind = Mega ex` applied — filters compose with the existing ones and persist in the URL:

Mega ex applied

greg19:
I'd remove trainers from the Pokemon types, mimicking the in-game filters. Including the behavior which couples these two: both filters are active if you select any of them, even if the other one is not active. The "select all" button can also be useful.

Screenshot_2026-08-31-10-06-48-911_jp pokemon pokemontcgp-edit

Screenshot_2026-08-31-10-07-01-046_jp pokemon pokemontcgp-edit

PS. The CSS for stage/item/trainer is a little bit awkward

diegomarnie:
On the CSS note for stage/kind/subtype — could you point to what specifically feels off? A couple of guesses so you don't have to type much:

- Missing labels above those toggle rows (users see "Basic / Stage 1 / Stage 2" with no header telling them what it filters).
- The bordered wrapper feeling underweight when a row only has 3-4 short items, next to rarity/type rows that are much fuller.
- Button spacing or widths inside the group.

Or something else? Want to make sure I target the right thing before I touch it.

greg19:
> On the CSS note for stage/kind/subtype — could you point to what specifically feels off?

The lack of margins and padding that filters with text buttons typically use. Your "ability" filter has them, but the other don't (I guess that's because you used a different React component with different defaults).

frontend/src/lib/filters.ts
```diff
@@ -9,11 +9,21 @@ export const sortByOptions = ['expansion-newest', 'rarity', 'type', 'recent'] as
 export const cardTypeOptions = cardTypes
 export const ownershipOptions = ['all', 'missing', 'registered'] as const
 export const tradingOptions = ['all', 'wanted', 'extra'] as const
+export const stageOptions = ['basic', 'stage1', 'stage2'] as const
+export const pokemonKindOptions = ['regular', 'ex', 'megaex'] as const
+export const trainerSubtypeOptions = ['item', 'supporter', 'tool', 'stadium'] as const
+export const abilityOptions = ['all', 'yes', 'no'] as const
+export const hpOptions = [0, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120, 130, 140, 150, 160, 170, 180, 190, 200, 210, 220, 240, 260, 280, 300, 350] as const
```
greg19: This seems fragile. If they add a pokemon with 230 HP we might forget to add this value here

diegomarnie:
Force-pushed with the following addressed:

1. Trainer removed from Pokémon types — cardTypeOptions no longer contains 'trainer'; that row only shows energy icons + dragon/colorless.

2. Two type filters coupled — Trainer sub-type filter (Item/Supporter/Tool/Stadium) now composes with the Pokémon type filter as a single OR: nothing selected → all cards; anything selected on either side → only those types on that side, empty side contributes zero. Matches the in-game behavior. Screenshot 2 below shows Item alone → only trainer items, zero Pokémon.

3. Select all button — on both the Pokémon and Trainer sub-type filters; toggles between "Select all" and "Clear" depending on state.

4. CSS margins/padding — ToggleGroupItem bumped from px-1 rounded-sm → px-3 py-1 rounded-md (matches TabsTrigger defaults), plus p-1 on the outer ToggleGroup. The stage/kind/subtype rows breathe like the Ability filter did. Also gave each filter block a tab-style label (Pokémon / Trainer / Stage / Kind) for visual consistency with Ability, and grouped Pokémon+Trainer via tighter spacing to communicate the coupling.

5. hpOptions fragility (inline review) — hpOptions and retreatOptions are now derived from allCards at load time instead of hardcoded. Side benefit: the previous hardcoded list had four options that don't exist in any card (260, 280, 300, 350) and was missing 230 that does — both fixed for free.

Captura de pantalla 2026-08-31 191658
Captura de pantalla 2026-08-31 192506

To be honest, something about the panel still feels slightly off to me too, but I couldn't put my finger on the exact thing. If you spot what it is, happy to iterate.

marcelpanse:
Personal TCG Pocket Col… 2026-09-01 at 5 28 08 PM

I improved the visual hierarchy by removing the labels within the boxes and adding a separate label which better separates the boxes. It saves space and makes it much more parsable. Also added a "more filters" link where I've put the more advanced filters you don't regularly use.

PR #1009, opened 2026-08-30
Remove setdetails

greg19:
Removes the unused `card.set_details` field. Reruns the scraper, which fixes the previously incorrect rarities of some cards in B4a, thus superseding #1006.


marcelpanse:
@greg19 if the rarities change then it affects the index which means it also needs an update script to adjust the card_amounts table and a cache version increase. I just pushed both. I'll keep the update script in the repo in case we need to do it again.

marcelpanse:
@greg19

marcelpanse:
migration script executed

Issue #123, opened 2025-02-10
Implement Figma design


marcelpanse:
Create some way of importing the cards from the spreadsheet

https://www.figma.com/design/kjLcHIVDRXfZmAZBSkPkkG/Pokemon-TCG-Pocket-Collection-Tracker?node-id=1-2&p=f

marcelpanse:
Implemented the dark design and menu header. Some other details still need to get added.

PR #1014, opened 2026-09-02
feat: add "What's new" changelog dialog

diegomarnie:
## Motivation

A lot of user-facing work has landed in the last two weeks (new cards, filter overhauls, missions, image manifest, deck stuff) and users who don't watch the repo have no way to notice any of it. Bringing it in-app closes that gap and doubles as a small trust signal.

## What this does

A new bell button in the header opens a "What's new" dialog listing recent `feat:` and `fix:` commits from `main`, grouped by date, with badges for type. A subtle red dot appears on the button when there's an unread entry since your last visit — tracked in `localStorage` by latest-seen SHA, so it clears the moment you open the dialog.

## Approach

- **Data**: `scripts/generate-changelog.ts` runs `git log -n 500` on `main`, filters to `feat:` / `fix:` (skips `chore`, `Rerun scraper`, merge commits), strips `(#123)` PR suffixes, caps at 60 entries, writes `frontend/public/changelog.json`.
- **UI**: `frontend/src/components/ChangelogDialog.tsx` fetches the JSON at runtime via react-query (1 h stale), groups by date, renders inside the same `Dialog` primitive already used for About Us.
- **Deploy**: the JSON is committed for this first pass so nothing changes in the build chain. If you'd like it regenerated on every deploy, one line in the `build` script does it (`tsx scripts/generate-changelog.ts && ...`) and the JSON can be gitignored. Happy to add that if you want it wired.

## What's intentionally out of scope

- Localization — commit messages are inline English. Adding a translation layer would be a follow-up.
- Filters / search / pagination — first pass keeps a plain scroll. Easy to add later if the list gets long.
- Rich content per entry — no summaries beyond the commit subject. Adding a `changelog:` trailer to commits could opt into human-friendly text, but wanted to see if the raw approach felt useful before asking for commit-message discipline.

## First-pass ergonomics I'm unsure about

- **Iconography**: bell felt closest to "notification"; other candidates were sparkle, megaphone, or something set-themed. Happy to swap.
- **Position**: placed between nav and the language dropdown; could also live inside the user dropdown menu.
- **Red dot**: currently only appears when there's a newer entry than the last-seen SHA. Considered showing the count too but felt noisy for a passive feed.

Screenshot coming in a follow-up comment.


diegomarnie:
Captura de pantalla 2026-09-02 134737

A few things worth calling out that aren't obvious from the diff:
- The scope prefix (collection, ui) is rendered in muted gray before the entry title, so you can scan by area without it dominating.
- Dates group entries and use the short month name in the viewer's locale, no code needed.
- Empty-day gaps are just skipped — no filler entries.
- The red-dot indicator on the bell doesn't appear here because I'd already opened the dialog once; it shows on first visit for a user, then clears when they open it.

Happy to grab a mobile view or a header-with-badge shot if useful.

greg19:
Why a dialog instead of a route? I personally would use dialogs only in intermediate steps of a longer workflow, where it encapsulates something (for example, the "choose card" in the trade matches form). Having a dialog has some disadvantages:
- It has to be loaded as an element in the browser even if it's never opened
- It can cause unexpected side effects. If I see correctly (I only briefly glanced over the code), you have a `useQuery` that's always enabled, so it makes an additional network request right when you open the page, even if the dialog is never opened.
- It wastes screen real estate. Especially on mobile it either looks as a separate page (then what's the point of dialog once again?) or is really small.
- It's easier to accidentally close it by clicking on the background.

marcelpanse:
I agree this would be nicer as a full page that we can also deeplink to. 

There needs to be a async network request after boot anyway because you need to fetch that json to see if there is anything newly released compared to your localstorage and if so there should be a numbered badge on the icon so you see something changed (and you click it to read it).

greg19:
We already have the community forum, which I believe is a perfect place for change announcements. Having change notes will disallow discussion (sometimes on announcement posts people are suggesting next steps or that it didn't work, ect.) and reduce the traffic on the community, discouraging communication from users to developers.

diegomarnie:
Thanks both — happy to iterate on all four points.

**On the technical asks — agree with all of them:**

- **Route instead of dialog**: convert the button into a link to `/changelog` (or similar), with the entry list as a full page. Solves the mobile real-estate problem and gives us a shareable URL.
- **Numbered badge**: swap the red dot for a count of entries newer than the last-seen SHA (localStorage). Only rendered when > 0.
- **Lazy fetch**: still need one request post-boot to compute the badge count, but we can gate it behind a tiny endpoint or just fetch the JSON once and cache it — no per-open re-fetches. Alternatively, only fetch when the user's `lastSeenSha` is missing or old; if we know the latest SHA some other way (e.g. included in the app bundle at build time), we skip the fetch entirely until they click.

Will rework the PR along those lines if the community-forum question below lands the right way.

**On the community-forum concern (@greg19):**

Fair point that discussion is a real cost — worth thinking through. My read is that the two serve different audiences, not the same one:

- The **forum** requires the user to opt in (know it exists, go there). That's fine for engaged users but the majority of visitors never leave `tcgpocketcollectiontracker.com` — for them, all the recent work (B4a, new filters, image manifest, missions) landed silently.
- An **in-app changelog** captures passive discovery: the user is already on the page, glances at the header, spots a badge, reads the summary. Not the same person who would post on the forum in the first place.

So the risk isn't so much "users read changelog instead of posting", it's "users who would never have posted anyway now know what shipped."

Concrete way to avoid canibalising forum discussion: **make the changelog explicitly point at the forum for discussion.** A single "Discuss recent updates on the forum →" link at the top or bottom of the page routes the users who *do* want to engage into the existing channel, and each entry stays short (no comment thread, no upvotes, no engagement surface competing with the forum).

Happy to drop the whole thing if you'd rather double down on forum visibility instead (e.g. surface the latest N forum posts in the header, or a "New in the community" strip on the overview page) — that's a different but equally valid answer to the same problem.

greg19:
I'd speculate that if a user is not engaged enough to regularly check the forum then they don't care about the patch notes.

We live in attention economy, where everything shouts to you "spend more time here", and some people want to opt out of it, which is completely fine. After all, this is just a mobile game (which already tries to grab as much of your attention as possible), maybe we don't need to prompt the users to spend additional 20 seconds (switching their attention and focus) every time we fix a flickering issue.

I think the forum is a perfect boundary, where dedicated people can "pull" information from, and people who just want to live their lives don't get "pushed" information they don't care about.

greg19:
If this lands, the user should be able to opt-out of receiving the badge, as it could be really annoying for some (which includes me, and it's not because I already track the changes). My (random) guess is that most of the users don't actually care, so I'd say that maybe even opt-in.

I don't spend my time reading patch notes of my browser or email client, which are (arguably) more important pieces of software I use than a random pokemon site. Please, treat the importance of the app accordingly to the value it brings to the users life. I feel this is about the respect of the users time and attention, which is scarce in these times.

marcelpanse:
Good points. We could also make it a separate page and only link it from the "my account" dropdown. So no icon/badge at all. Then you also only have to fetch the json on the page itself. It will be a bit buried though, so I don't many people will actually go and see it.

diegomarnie:
Reworked to match your proposal. Pushed 2198cc3e:

- Bell button gone from the header entirely.
- New `/changelog` route (lazy-loaded page), linked from the "My account" dropdown as "What's new".
- Small unread-count badge lives inside the dropdown item — only visible once the user opens their account menu, so no push surface on the header at all.
- Zero runtime fetch: the JSON is bundled at build time and imported statically, so the badge count and the page render from the same payload. Nothing hits the network on app boot.
- The `/changelog` page opens with a "Discuss on the community forum →" link at the top, routing engaged readers toward the existing channel.

Left the mobile hamburger untouched — @marcelpanse's proposal specifically named the my-account dropdown. Happy to mirror the entry there for form-factor consistency if you'd like.

marcelpanse:
some feedback on the design, it looks messy because the texts are not aligned: 

Personal TCG Pocket Col… 2026-09-06 at 10 35 09 AM

make sure the pills in front have the same width, and maybe put the tag labels at the end, so each description start is aligned. 

Also, below the "Recent changes to the tracker" sentence, put another sentence about that this is open source and if you want to help you go to the github repo. Where "github repo" is a link.

diegomarnie:
Applied both changes in 0e4f6fc6:

- **Type pills moved to the end of each row.** Every description now starts in the same column, so the ragged left edge from the previous screenshot is gone. Titles that wrap still keep the pill aligned on the first line.
- **New line under the subheader** mentioning the project is open source, with `github repo` linking to https://github.com/marcelpanse/tcg-pocket-collection-tracker.

Copy is unlocalized like the rest of the page — happy to add locale keys if you want it translated before merge.

marcelpanse:
That's still not aligned. Also I would appreciate screenshots when you do something with UI so I can spot it immediately. Now I have to pull your code, start it and then realize it's not good. See below, the "collection", "ui" pills break the alignment still. 

image

diegomarnie:
Good catch — the `collection` / `ui` prefixes were still in front of the title, which broke the left edge. Reworked the row layout so **everything secondary sits on the right**: title starts on the same column across every entry, and the scope tag (when present) is a small muted label just before the type pill.

Pushed as f03fec94.

Screenshot from `/#/changelog` on the current branch:

Captura de pantalla 2026-09-07 220749

And point taken about screenshots — I'll attach one alongside any UI-touching push from now on.

PR #1017, opened 2026-09-07
fix(scraper): refresh 16 stale set logos + always re-fetch on scrape

diegomarnie:
## Summary

`scripts/download-packs.js` used to skip downloads whenever a logo already existed on disk (`if (!fs.existsSync(dest))`), so any logo Limitless replaced after our initial fetch stayed frozen indefinitely. The most visible symptom was **B4a** — Limitless swapped its early placeholder for the real "Team Rocket's Ambition" logo months ago, but we kept serving the placeholder.

The script now downloads every logo on each run and only writes when the bytes differ, so future replacements are picked up automatically on the next `pnpm scraper`.

Running it now refreshed **16 logos**: `A1, A1a, A2, A2a, A2b, A3, A3a, A3b, A4, A4a, A4b, B1, B1a, B3a, B4a, P-A`. `B2..B4` and `B1a` already matched Limitless. The write-only-on-diff check keeps git tidy for the ones that don't change.

## Also in this PR

- **`scripts/sql/generate-cards-table.ts`**: rename `tradeableExpansions` → `tradableExpansions` to match #1015. Without this the SQL step of the scraper pipeline crashes with `SyntaxError: The requested module '../../frontend/src/lib/CardsDB' does not provide an export named 'tradeableExpansions'` and the bulk-insert never regenerates.
- **`frontend/assets/cards.json`**: minor Limitless updates picked up in the same scraper run — a handful of `alternate_versions` entries for Team Rocket cards, plus corrected wording on several P-B card effects (curly quotes, missing final periods, "Grass Energy" → "[G] Energy").

## Not included

There's a related client-side issue that causes the same silent-staleness on the user side: `frontend/public/service-worker.js` is cache-first over `images/` with no revalidation, so users who already visited the site keep seeing whatever they cached the first time — even after we ship new binaries. I confirmed it by clearing the SW in Firefox and the refreshed logos then appeared instantly. That fix is a separate PR to keep this diff focused on the scrape side.

## Test plan

- [ ] `pnpm scraper` runs cleanly through all steps (including the SQL generation that used to crash).
- [ ] Re-running `pnpm scraper` on a clean tree produces no diff (idempotent — proves the write-only-on-diff guard works).
- [ ] Visually inspect the 16 refreshed set logos in the sets picker — each looks like the current Limitless page.


marcelpanse:
The PR removes HR versions of images and replaces them with low res version. Some people source HR images from other sources (like drive), so they shouldn't be overwritten by a low-res scraped version.

diegomarnie:
Fair point — overwriting a manually-curated HR version with a Limitless placeholder is exactly the wrong direction, and this PR was doing that as a side effect of the idempotence fix. Closing.

Splitting what's left into two smaller pieces:

- The one line that was clearly a bug regardless — the `tradeableExpansions` import in `generate-cards-table.ts` still referencing the pre-#1015 name — is in #1026 on its own.
- The `download-packs.js` idempotence tweak I'll hold off on until there's a way for it to detect and preserve HR overrides (bytes larger than the Limitless copy, a manual "don't touch" manifest, or just a hand-curated whitelist per file). Happy to take direction on which shape you'd prefer before opening anything.

Sorry for the noise.

PR #1024, opened 2026-09-07
fix(missions): scrape full card pools via pokemon-zone API

diegomarnie:
## Summary

Reported on the community forum: several themed-collection missions with large card pools (**Team Rocket's Pokémon**, **A Sweets Party of Epic Proportions**, and more) were only tracking progress for a single card instead of every eligible one. The forum comment:

> Some missions that allow a large number of cards to be used, e.g. "Team Rocket's Pokémon" or "A Sweets Party of Epic Proportions" don't track all cards that can be used, but just one card.
> Especially "A Sweets Party of Epic Proportions" worked some months ago as intended, but is now broken.

The regression was introduced in #914 when I rewrote the missions scraper to pokemon-zone. That scraper reads the mission listing HTML, which only shows one representative card + a "+N Cards" counter for pool missions. My fallback resolved that placeholder through the visible card's `alternate_versions` in cards.json — fine for cross-set reprints but wrong for missions whose pool is a genuinely different set of cards (all Team Rocket's Pokémon, all Alcremie, …). Those collapsed to the one visible card.

## Fix

pokemon-zone exposes the full pool through the same private API its mission modal uses:

```
GET /api/missions/themed-collections/{missionId}/
```

Every mission-card in the listing carries a `data-mission-id`. The rewritten scraper now:

- Fetches `/api/game/card-data/` once to build a map from pokemon-zone card ids (e.g. `PK_10_021240_00`) to every card_id in cards.json. Cards that live in multiple expansions map to all of them (Bulbasaur is both A1-1 and A4b-1, both valid options).
- Reads the expansion listing for name / reward / secret status like before.
- Calls the mission API for each mission and resolves every `targetId` in every frame's criterion to card_ids. Handles the three `targetType` variants pokemon-zone uses — `CARD_ID`, `POKEMON_CARD_ID`, `TRAINERS_CARD_ID` — and warns on anything else so a new game version doesn't silently produce empty options again.
- Falls back to the visible-figure list when the API returns 500 on a mission id it doesn't recognise.
- Skips P-A / P-B (promo sets have no themed-collections page — 404).

## Results across the 21 regenerated JSONs

- **Team Rocket's Pokémon** (B4a): 1 → 46 options.
- **A Sweets Party of Epic Proportions** (A3b): 2 → 11 options.
- **ex Collection** (A4b): 3 → 166.
- **Shiny Radiance** (B2b): 2 → 29.
- **A Variety of Pokémon Abilities** (A2): 1 → 22.
- **Collect 1 Mega Evolution Pokémon ex card** (B1): 4 → 18.
- Zero missions with empty options across 445 total.

Some smaller missions come out with a couple fewer options per slot where the previous scraper had over-included alternates the game doesn't actually accept — a correction, not a regression. Example: "First Choice" (A1) had 5 options per starter slot but the game only accepts 3 (base + regular ex + parallel).

## Test plan

- [ ] `pnpm tsx scripts/scrape-missions.ts` runs cleanly and re-produces the same JSONs on a clean tree (idempotent).
- [ ] Open the app → Missions view → check "Team Rocket's Pokémon" and "A Sweets Party of Epic Proportions" show the full pool of eligible cards.
- [ ] Sanity-check any other flagged mission from the community forum.


marcelpanse:
when I rerun mission scraper it produces differences with this PR. For example:
```
       {
-        "options": ["A1-168", "A1-240"],
+        "options": ["A1-168"],
         "amount": 1
       },
```

four of these in total, why is that?

diegomarnie:
Just re-ran the scraper locally against `main` — after `pnpm exec biome format` the working tree is clean (no diff vs. what this PR ships). So the output is reproducible from pokemon-zone.

On the specific `["A1-168", "A1-240"]` → `["A1-168"]` diff: those two ids are the two variants of the same Nidoqueen — regular ◊◊◊ (168) and foil ☆ (240). They appear paired because a single `PK_id` in pokemon-zone's card catalog carries **both** entries in its `expansionCollectionNumbers`, and `resolveTargetIds` unions them. Whenever a mission mentions "Nidoqueen (Lovestrike)" both variants count, which matches in-game behaviour.

There are 24 pairs like this in `A1-missions.json` alone (Charmander/Bulbasaur/Pikachu/… regular + foil), all of the same shape. If Marcel's rerun collapsed 4 of them, my guess is a transient pokemon-zone response — either the mission-detail API 500'd on one attempt and dropped through to the visible-slots fallback (which only sees the figure shown in the card, i.e. the regular variant), or curl's `--retry 3` didn't kick in fast enough. It hasn't reproduced on my end.

If you want, paste the 4 diffs and I'll check each one — `A Day with Family` (line 88) and `Complete the Kanto Pokédex!` (line 492) are the only A1 missions that carry `168+240` at all, so I can pinpoint which frame is affected.

marcelpanse:
Now I can't run it at all anymore

```
➜  tcg-pocket-collection-tracker git:(fix-missions-scraper) tsx scripts/scrape-missions.ts
Fetching pokemon-zone card catalog...
node:internal/errors:985
  const err = new Error(message);
              ^

Error: Command failed: curl -sL --compressed --fail --retry 3 --retry-delay 5 -H User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0 Safari/537.36 -H Accept: application/json -H Accept-Language: en-US,en;q=0.9 -H Sec-Fetch-Dest: document -H Sec-Fetch-Mode: navigate -H Upgrade-Insecure-Requests: 1 https://www.pokemon-zone.com/api/game/card-data/
    at genericNodeError (node:internal/errors:985:15)
    at wrappedFn (node:internal/errors:539:14)
    at checkExecSyncError (node:child_process:925:11)
    at execFileSync (node:child_process:961:15)
    at fetchWithCurl (/Users/marcelpanse/Documents/Sources/tcg-pocket-collection-tracker/scripts/scrape-missions.ts:42:10)
    at fetchJSON (/Users/marcelpanse/Documents/Sources/tcg-pocket-collection-tracker/scripts/scrape-missions.ts:50:21)
    at buildCardIdIndex (/Users/marcelpanse/Documents/Sources/tcg-pocket-collection-tracker/scripts/scrape-missions.ts:96:20)
    at main (/Users/marcelpanse/Documents/Sources/tcg-pocket-collection-tracker/scripts/scrape-missions.ts:262:17)
    at <anonymous> (/Users/marcelpanse/Documents/Sources/tcg-pocket-collection-tracker/scripts/scrape-missions.ts:283:1) {
  status: 56,
  signal: null,
  output: [ null, '', '' ],
  pid: 49578,
  stdout: '',
  stderr: ''
}

Node.js v24.15.0
```

PR #1029, opened 2026-09-15
Add buildable decks filter

### LINKED ISSUE #867: Filter decks you are able to build
I'd be cool to have a filter on the decks page for showing only decks you are able to build.

It's a little bit complicated, because we need to join the `card_amounts` table with `decks` and write some logic for calculating the number of cards the user has for the deck. We would have to introduce a `alternate_version(card_id, card_id)` (or `representative_version(card_id, card_id)` that would correspond to `card.alternate_versions[0]`) to do the alternate cards substitutions on the DB side, and remember to keep this table up to date.

Another thing to investigate is whether the JS api allows for this kind of complicated logic, or we would need to switch to SQL for deck fetching.

greg19:
Closes #867.


greg19:
@marcelpanse Can you apply the DB changes?

PR #1031, opened 2026-09-19
Filter buildable decks client-side instead of in the database

marcelpanse:
Alternative implementation of the "Decks I can build" filter from #1029, which this builds on — the first two commits are @greg19's, unchanged. Supersedes the `get_buildable_decks` SQL function with client-side filtering.

## Why

The buildable predicate needs two things: the user's full collection and the `alternate_versions` groups from `cards.json`. Both already live in the browser. Doing the filtering server-side meant mirroring `alternate_versions[0]` into a new `cards_list.deckbuilding_id` column, which brought along a migration, a `bulk-insert.sql` regeneration, a `schema.sql` re-dump, and a second copy of the "is this deck buildable" rule that could drift from the frontend's.

Net diff vs `main` drops from 8 files / ~3800 lines to 4 files / ~130.

## What changed

- **Removed:** `scripts/sql/get-buildable-decks.sql`, `scripts/sql/migrations/2026-09-add-deckbuilding-id.sql`, and the `deckbuilding_id` changes to `generate-cards-table.ts` / `output/bulk-insert.sql`. No schema change, so nothing to apply by hand in the SQL editor.
- **Kept:** the `cards.json` fix adding promo cards to their `alternate_versions` groups. That's an independent data correction and the client-side predicate reads it directly.
- **`deckService.ts`:** `getAllPublicDecks()` pages through `public_decks` in 1000-row chunks (PostgREST caps rows per request); `selectBuildableDecks()` applies the energy filter, the buildable predicate, ordering and pagination locally.
- **`useDeck.ts`:** `useDecksSearch` keeps the paginated query for the three non-buildable tabs and otherwise uses a single long-lived `['decks','all-public']` query plus a `useMemo`. Now keyed off `useCollection` rather than `account.collection_last_updated`, which also fixes the duplicate fetch on a cold load of `#/decks?buildable=true`.
- **`Decks.tsx`:** `needsLogin` is checked before `isLoading`, otherwise anonymous users would sit on a spinner instead of the sign-in prompt.

Buildability now goes through `getMissingCardsCount`, the same helper `DeckView` uses for its missing-cards count, so the list and the badge can't disagree — including on stale `internal_id`s left behind by a rarity fix, which both now treat identically.

## Trade-off

The first toggle of the checkbox downloads every public deck (~300B/row, so roughly 1.5 MB uncompressed / ~400 KB gzipped at 5k decks), after which paging, re-sorting and energy toggles are instant and need no further requests. The SQL function scanned every public deck on *every* page change, so this is a bandwidth-for-latency swap rather than a straight regression — but it does scale with the deck table. If public decks grow past ~20k, the alternative is a chunked scan that fetches in sort order and stops once a page is filled, at the cost of the exact result count and numbered pages.

Worth a look before merge: actual payload size against production deck counts.


marcelpanse:
@greg19 I branched of your branch for the deck builder feature. Can you take a look at this. I made it client side, which makes it a lot simpler (no sql function, no db migration). Can you try it and see if you agree with this approach instead?

greg19:
Eventually we will want to have deckbuilding serverside for #969. It's more interesting and it was the next thing I wanted to tackle. Also it's a good rule of thumb to do computation before network transfers. If a person with a small collection (let's say they can only build 10 decks on the site) were to check this option, then we would need to transfer the *whole* table to them. As your AI pointed out, this approach would require much more engineering if the table grows. So I don't agree.

PS. I think pointing out that this "drops the diff size" while counting auto-generated files is a dishonest manipulation. Mine is actually smaller when you discard auto-generated files (120 vs 126). But I don't even know what's the point in doing that comparison.

marcelpanse:
I don't care about the diff size. Only about complexity. The full decks table is not even 200 rows, it would be different if it was 5000+ but I suspect it won't grow to that size anytime soon. So regarding it is so few records it's just simpler to do it like this.

greg19:
Okay, but we will still need the DB migration if we want the trading feature.

marcelpanse:
migration script executed and schema/cards db updated.

